PR #65, opened 2023-01-12
fw_update: add golioth_fw_update_init_with_config()

ncmiller:
This new API will allow the user to specify the name of the main firmware package, in case something other than "main" is desired.

Created a new golioth_fw_update_config_t struct which is passed into the new API, which may be helpful to maintain backward compatibility in case more user configuration parameters are added in the future.

Signed-off-by: Nick Miller <nick@golioth.io>


szczys (approved):
LGTM!

PR #81, opened 2023-02-01
golioth_basics: change 'int' to 'double' in 'multiply'

mniestroj:
All numbers in cJSON library are first assigned to `valuedouble` and then
converted to `valueint` for simplified access when they are expected to be
integers on the application side.

Use `valuedouble`, so that `1.5` can be multiplied by `3`, resulting in `4.5`:

```sh
$ golioth call -d nrf52 multiply 1.5 3
{'value': 4.5}
```

### DIFF
--- a/examples/common/golioth_basics.c
+++ b/examples/common/golioth_basics.c
@@ -46,9 +46,9 @@ static golioth_rpc_status_t on_multiply(
     if (cJSON_GetArraySize(params) != 2) {
         return RPC_INVALID_ARGUMENT;
     }
-    int a = cJSON_GetArrayItem(params, 0)->valueint;
-    int b = cJSON_GetArrayItem(params, 1)->valueint;
-    snprintf((char*)detail, detail_size, "{ \"value\": %d }", a * b);
+    double a = cJSON_GetArrayItem(params, 0)->valuedouble;
+    double b = cJSON_GetArrayItem(params, 1)->valuedouble;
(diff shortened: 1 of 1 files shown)


ncmiller (approved):

examples/common/golioth_basics.c
```diff
@@ -46,9 +46,9 @@ static golioth_rpc_status_t on_multiply(
     if (cJSON_GetArraySize(params) != 2) {
         return RPC_INVALID_ARGUMENT;
     }
-    int a = cJSON_GetArrayItem(params, 0)->valueint;
-    int b = cJSON_GetArrayItem(params, 1)->valueint;
-    snprintf((char*)detail, detail_size, "{ \"value\": %d }", a * b);
+    double a = cJSON_GetArrayItem(params, 0)->valuedouble;
+    double b = cJSON_GetArrayItem(params, 1)->valuedouble;
+    snprintf((char*)detail, detail_size, "{ \"value\": %lf }", a * b);
```
mniestroj, line 51: In some configurations `snprintf()` does not handle printing floats. This is true for example with minimal picolibc configurations. In that case following string was generated: `{ "value": *float* }`, which is invalid and is not gracefully handled by cloud (obviously).

So I wonder if we should abandon this change instead and rely on integers.

@ncmiller thoughts?

ncmiller: Yeah, I think it probably makes sense to abandon this change, so we can be compatible with minimal libc implementations like `picolibc`.

PR #83, opened 2023-02-08
submodule: change heatshrink URL to golioth/heatshrink

ncmiller:
The submodule now points to the Golioth fork of
`atomicobject/heatshrink`.

The fork is required in order to change configuration in
heatshrink_config.h, which will happen in subsequent commits.

For local repo clones, you will need to update with:

```
git submodule sync --recursive
git submodule update --recursive
```

Signed-off-by: Nick Miller <nick@golioth.io>

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,4 +12,4 @@
 	url = https://github.com/DaveGamble/cJSON
 [submodule "external/heatshrink"]
 	path = external/heatshrink
-	url = https://github.com/atomicobject/heatshrink.git
+	url = https://github.com/golioth/heatshrink.git


mniestroj (commented):
Right now we were using `master` branch of heatshrink, which doesn't have any update for the last 7 years. There is also `develop` branch, which has some changes made about 2 years ago. Looking at those changes there are mainly some improvements (minor ones, not affecting the algorithm or anything big), fixed typos, suppressed compiler warnings, etc. It is good idea to pull them in, especially because of this commit: https://github.com/atomicobject/heatshrink/commit/0587f557ebbe, which allows to override configuration, which is basically what we need. Looking at the `heatshrink_decoder.c` (which is the only module we use) there are minor changes, mostly related to suppressing clang warnings. This is something we want anyway, due to clang getting more likely to be supported over time in embedded space.

So what I propose is simply drop forking heatshrink repo and just bump commit from `master` to `develop`. Then we will be able to override configuration however we want.

ncmiller:
> So what I propose is simply drop forking heatshrink repo and just bump commit from master to develop. Then we will be able to override configuration however we want.

Good call. I didn't even think to check for other branches, so I had assumed `master` was the latest. I'll close this PR.

PR #84, opened 2023-02-08
fw_update: use static heatshrink decoder

ncmiller:
* heatshrink: update to latest develop branch 
* fw_block_processor: use static heatshrink decoder

### DIFF
--- a/src/golioth_fw_update.c
+++ b/src/golioth_fw_update.c
@@ -43,7 +43,6 @@ static golioth_status_t download_and_write_flash(void) {
     GLTH_LOGI(TAG, "Download took %" PRIu64 " ms", golioth_sys_now_ms() - start_time_ms);
 
     fw_block_processor_log_results(&fw_block_processor);
-    fw_block_processor_deinit(&fw_block_processor);
 
     fw_update_post_download();
 
--- a/examples/modus_toolbox/golioth_basics/golioth_app/Makefile
+++ b/examples/modus_toolbox/golioth_basics/golioth_app/Makefile
@@ -224,6 +224,7 @@ DEFINES+=APP_VERSION_MAJOR=$(APP_VERSION_MAJOR) \
          USER_APP_START=$(PRIMARY_IMG_START) \
          USER_APP_SIZE=$(SLOT_SIZE)
 DEFINES+=CONFIG_GOLIOTH_USER_CONFIG_INCLUDE='"golioth_user_config.h"'
+DEFINES+=HEATSHRINK_DYNAMIC_ALLOC=0
 
 # Add defines to enable usage of external flash for secondary or both images (XIP)
 ifeq ($(USE_EXTERNAL_FLASH), 1)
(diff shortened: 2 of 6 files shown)


src/golioth_fw_update.c
```diff
@@ -103,52 +100,50 @@ static golioth_status_t download_and_write_flash(void) {
...
+                heatshrink_decoder_sink(&_hsd, _ota_block_buffer, block_nbytes, &sunk);
+        if (sink_res != HSDR_SINK_OK) {
+            GLTH_LOGE(TAG, "sink error: %d, sunk = %" PRIu32, sink_res, (uint32_t)sunk);
+        }
 
-            // Pull the uncompressed data out of the decoder
-            HSD_poll_res pres;
-            uint8_t decode_buffer[HEATSHRINK_DECODE_BUFFER_SIZE];
-            do {
-                size_t poll_sz = 0;
```
mniestroj, line 119: That area of changes is challenging to review. Looking at the old code (before the change) there are 18 (!) local variables already at this point (`size_t poll_sz = 0;` is the 18th declaration). I'll list them just to point how many of them there are:
- `int32_t main_size`
- `heatshrink_decoder* hsd`
- `block_latency_stats_t stats`
- `uint64_t start_time_ms`
- `size_t nblocks`
- `size_t bytes_written`
- `size_t i`
- `size_t block_nbytes`
- `bool is_last_block`
- `uint64_t block_start_ms`
- `golioth_status_t status`
- `uint64_t block_latency_ms`
- `const float alpha`
- `size_t sunk`
- `HSD_sink_res sink_res`
- `HSD_poll_res pres`
- `uint8_t decode_buffer[HEATSHRINK_DECODE_BUFFER_SIZE]`
- `size_t poll_sz`

Additionally there are 3 global variables that this function is operating on:
- `static golioth_client_t _client`
- `static const golioth_ota_component_t* _main_component`
- `static uint8_t _ota_block_buffer[GOLIOTH_OTA_BLOCKSIZE + 1]`

This is overwhelming to keep so many variables in mind when analyzing that function. It should definitely be split into smaller entities (not saying it should be done now, just wanted to express my thoughts about it now). When there are so many variables, then I totally agree with the statement in commit body: "dynamic allocation of the decoder makes the SDK fw_update code more complicated and prone to leaks". That is true, because it is 1 of 18 variables managed in this function, which gives it only about 5% of developer attention.

A good quote from https://www.kernel.org/doc/html/v4.10/process/coding-style.html#functions:
> Another measure of the function is the number of local variables. They shouldn’t exceed 5-10, or you’re doing something wrong. Re-think the function, and split it into smaller pieces. A human brain can generally easily keep track of about 7 different things, anything more and it gets confused. You know you’re brilliant, but maybe you’d like to understand what you did 2 weeks from now.

One reason why I personally like to put all declarations at the top of the function is an easy way to track number of variables. The decision to do so is kind of "code style" issue, but the reason is actually not only "good looking code" (that is personal really) but rather a good "measure" of number of local variables. It is simply easy to spot that something is wrong (too many local variables) just by looking at the top of function body.

This comment is not that relevant to future changes or state of the repo, but in general I wanted to point out the problem. I'll continue review anyway.

ncmiller: Comments above have been addressed by https://github.com/golioth/golioth-firmware-sdk/pull/87 and https://github.com/golioth/golioth-firmware-sdk/pull/88.

mniestroj (approved):

PR #88, opened 2023-02-10
fw_update: refactor download_and_write_flash() to simplify and reduce variables

ncmiller:
A series of commits to simplify the logic in `golioth_fw_update.c:download_and_write_flash()`, in preparation for adding an OTA delta feature in the near future, and to address [this](https://github.com/golioth/golioth-firmware-sdk/pull/84#discussion_r1102812321) comment about there being too many variables to keep track of in `download_and_write_flash`.

This series of commits does not change the observable behavior of `download_and_write_flash()`, and thus can be considered a "refactor" of the internal implementation of that function.

The primary structural change introduced in this sequence of commits is to split the processing of individual OTA blocks into three separate stages, where the output of one stage is connected to the input of the next stage via function pointer:

```
download() -> decompress() -> handle_block()
```

This decouples the individual processing stages from each other, and makes it easy to insert a new `patch` stage in future commits, as part of the OTA delta feature implementation.


src/golioth_block_processor.c
```diff
@@ -0,0 +1,202 @@
...
+}
+
+static golioth_status_t handle_block(const uint8_t* in_data, size_t in_data_size, void* arg) {
+    handle_block_ctx_t* ctx = (handle_block_ctx_t*)arg;
+
+    golioth_status_t status = fw_update_handle_block(
+            in_data,
+            in_data_size,
+            ctx->bytes_handled,  // offset
+            ctx->component_size);
```
mniestroj, line 109: This component (`golioth_block_processor.c`) is still FW related in this part.
Because of that, I would rather rename it to `golioth_fw_block_processor.c`.

mniestroj:
Thanks a lot of this refactoring, it is in much better shape now. The fact that all the processing stages are separate now makes this code much easier to read and understand. As a side effect we are really close to supporting decompression of any type of artifacts/blobs downloaded from Golioth cloud, not just FW artifacts. The only comment I have right now is about the name of the file `golioth_block_processor.c` to make it clear it is FW related (still at this point). Otherwise LGTM!

mniestroj (approved):

PR #93, opened 2023-02-19
.github: deploy new OTA release from self-hosted runner

ncmiller:
Previously, we relied on the repo secret GOLIOTH_API_KEY to be able to deploy the OTA release from GitHub Actions in the cloud.

This is problematic in the case where there are multiple self-hosted runners, each which might be using a different Golioth project with different API keys.

To resolve, we will deploy the OTA release from the self-hosted runner, where the runner-specific API key is already known and present in the file $HOME/credentials_esp32s3.yml.

This will allow for deleting the repo secret GOLIOTH_API_KEY.

### DIFF
--- a/.github/workflows/test_esp32s3.yml
+++ b/.github/workflows/test_esp32s3.yml
@@ -21,13 +21,6 @@ jobs:
       with:
         python-version: 3.x
         architecture: 'x64'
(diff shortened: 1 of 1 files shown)


szczys (approved):
LGTM!

PR #94, opened 2023-02-20
examples,nvs: increase max num chars for wifi params to 32

ncmiller:
Before this change, the max number of characters that could be used
for the WiFi SSID and password was 31.

WPA2-PSK security allows for a PSK up to 32 characters in length,
so we should be able to accommodate PSKs of that length.


szczys (approved):
I tested this using the golioth_basics example on an ESP32 and it works.

PR #102, opened 2023-03-05
README: clarify the support for OTA on Linux

### LINKED ISSUE #101: Status of OTA on Linux
The "[Features Supported by Platform](https://github.com/golioth/golioth-firmware-sdk#features-supported-by-platform)" table in the README states that `OTA FW Update` & `Sign+Verify OTA images` are coming soon, while `OTA FW compression` & `OTA FW delta update` are supported.

Is that accurate? I would assume that until `OTA FW Update` is implemented, no other OTA-related feature can be supported.

Thanks!

ncmiller:
Before this change, the features table indicated that Linux "OTA FW Update" was coming :soon:. However, other OTA features like compression and delta updates were already marked as supported, which was confusing, since one would expect the baseline FW update functionality to be a prerequisite for the other features.

This marks "OTA FW Update" as supported for Linux, but with a note that explains it is only partially implemented.

Fixes #101

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ the :x:'s indicate a feature that is not planned (or not applicable).
 
 | Feature | ESP-IDF | ModusToolbox | Linux |
 | --- | --- | --- | --- |
-| OTA FW Update | :heavy_check_mark: | :heavy_check_mark: | :soon: |
+| OTA FW Update | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: <sup>1</sup> |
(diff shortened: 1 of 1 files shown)


beriberikix:
Any idea why the Doxygen build is failing?

ncmiller:
> Any idea why the Doxygen build is failing?

I think so. Attempting to fix in https://github.com/golioth/golioth-firmware-sdk/pull/103.

mniestroj (approved):

PR #103, opened 2023-03-06
.github: use "on: pull_request_target" for doxygen

ncmiller:
For PRs from forked pull requests, the `firebase-doxygen-pr.yml` workflow will fail with the following error:

```
/home/runner/work/_actions/FirebaseExtended/action-hosting-deploy/v0/bin/action.min.js:274
        throw new Error(`Input required and not supplied: ${name}`);
        ^
```

This happens because the workflow does not have access to the repository secrets from the fork.

This is being resolved by changing the workflow to trigger on `pull_request_target`, which means it runs in the context of the base of the pull request, which has access to the repo secrets.

The workflow still requires approval from a Golioth team member to run, so from a security perspective, it will be as secure as it was before this commit.

More information is available here:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

### DIFF
--- a/.github/workflows/firebase-doxygen-pr.yml
+++ b/.github/workflows/firebase-doxygen-pr.yml
@@ -1,5 +1,5 @@
 name: Deploy Doxygen to Firebase Hosting on PR
-'on': pull_request
+'on': pull_request_target
 jobs:
   deploy_doxygen_dev:
     runs-on: ubuntu-latest


mniestroj (approved):

mniestroj:
Looks like right now documentation from the tip of `main` is built, instead of the PR. This means that there is no error when building (although to be tested/confirmed), but instead any documentation updates are not built automatically. This most likely affects all the PRs created from this repo, so either we'll find a proper solution to this or we need to revert this patch.

PR #110, opened 2023-03-09
examples: zephyr: support nrf52840dk_nrf52840

mniestroj:
Those helpers are ported from Zephyr SDK. As a result it is enough to call
`net_connect()` in sample's `main()` function in order to connect using
supported network technology and obtain IP address.

Add overlays for use with nRF52840 DK with ESP32 (using ESP-AT firmware).


szczys:
LGTM!

I tested installing and building for the nRF52840 and found all parts of `golioth_basics` to be working except for OTA (which is not yet implemented).

PR #117, opened 2023-06-01
unit_tests: Add RPC unit test

sam-golioth:
Adds unit tests for the RPC module.

Also adds a set of fake functions for golioth_coap_client and golioth_time.

Also cleans up the Unit Test CMakeLists.txt and adds a convenience function for declaring unit tests.

### DIFF
--- a/src/golioth_rpc.c
+++ b/src/golioth_rpc.c
@@ -17,7 +17,7 @@ LOG_TAG_DEFINE(golioth_rpc);
 // Request:
 //
 // {
-//      "id": "id_string,
+//      "id": "id_string",
 //      "method": "method_name_string",
 //      "params": [...]
 // }
@@ -127,6 +127,7 @@ static void on_rpc(
         }
     }
     if (!method_found) {
+        GLTH_LOGW(TAG, "Method %s not registered", rpc_method->valuestring);
         golioth_rpc_ack_internal(client, call_id, RPC_UNAVAILABLE, detail, 0);
     }
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -8,25 +8,66 @@ set(CMAKE_C_FLAGS "-fprofile-arcs -ftest-coverage")
 set(CMAKE_EXE_LINKER_FLAGS "-fprofile-arcs -ftest-coverage -coverage")
 
 set(repo_root ..)
+enable_testing()
+
+# Build Unity
+
 add_library(unity STATIC ${repo_root}/external/unity/src/unity.c)
 target_compile_definitions(unity PUBLIC -DUNITY_INCLUDE_CONFIG_H)
 target_include_directories(unity PUBLIC ${repo_root}/external/unity/src .)
 
+# Build FFF
+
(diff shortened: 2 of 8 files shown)


test/CMakeLists.txt
```diff
@@ -8,25 +8,61 @@ set(CMAKE_C_FLAGS "-fprofile-arcs -ftest-coverage")
...
+    fakes/golioth_time_fake.c
+)
+target_include_directories(test_golioth_rpc PRIVATE
+    ${repo_root}/external/libcoap/include
+    ${repo_root}/port/esp_idf/libcoap/include
+    ${repo_root}/port/linux
 )
-target_link_libraries(${test_name} unity fff)
-add_test(NAME ${test_name} COMMAND "./${test_name}")
+target_link_libraries(test_golioth_rpc cjson)
```
mniestroj, line 68: There is no newline at the end of this file as well as some other files. Would be good to clean that up to remove noise in future commits. 

We use pre-commit to enforce that in CI for Zephyr SDK: https://github.com/golioth/golioth-zephyr-sdk/blob/main/.github/workflows/pre-commit.yml with https://github.com/golioth/golioth-zephyr-sdk/blob/main/.pre-commit-config.yaml configuration file. Maybe we should add such github action to this repo as well?

src/golioth_rpc.c
```diff
@@ -127,6 +127,7 @@ static void on_rpc(
         }
     }
     if (!method_found) {
+        GLTH_LOGW(TAG, "Method %s not registered", rpc_method->valuestring);
```
mniestroj, line 130: This is not related to "Add RPC unit test", so would preferably split it to separate commit.

src/golioth_rpc.c
```diff
@@ -17,7 +17,7 @@ LOG_TAG_DEFINE(golioth_rpc);
 // Request:
 //
 // {
-//      "id": "id_string,
+//      "id": "id_string",
```
mniestroj, line 20: This is not related to "Add RPC unit test", so would preferably split it to separate commit.

test/CMakeLists.txt
```diff
@@ -8,25 +8,61 @@ set(CMAKE_C_FLAGS "-fprofile-arcs -ftest-coverage")
...
+
+golioth_unit_test(test_golioth_rpc
+    test_golioth_rpc.c
+    fakes/golioth_coap_client_fake.c
+    fakes/golioth_time_fake.c
+)
+target_include_directories(test_golioth_rpc PRIVATE
+    ${repo_root}/external/libcoap/include
+    ${repo_root}/port/esp_idf/libcoap/include
+    ${repo_root}/port/linux
```
mniestroj, line 66: Why do we include both `esp_idf` as well as `linux` port include directories? Shouldn't single port directory be included?

test/CMakeLists.txt
sam-golioth: I'll open another PR for adding the pre-commit hook

mniestroj (approved):

PR #118, opened 2023-06-02
scripts: only suggest running clang-format on files that don't match …

sam-golioth:
…coding standard


szczys:
I'm not quite sure what this does so I haven't approved it. However I was wondering about commit message formatting. Do you think we should adhere to gitlint rules (and commit signing) on this repo?

```
➜ gitlint
1: T1 Title exceeds max length (84>72): "scripts: only suggest running clang-format on files that don't match coding standard"
3: B6 Body message is missing
```

PR #128, opened 2023-07-17
config: commonize Kconfig options

sam-golioth:
Use a common Kconfig file for Zephyr and ESP-IDF

The diff doesn't look great here, but there aren't any changes to any of the options, just moving them from `port/*` to `src/`


szczys (approved):
LGTM

PR #129, opened 2023-07-19
Zephyr: examples: add hello

szczys:
* add samples common support for hardcoded credentials
* add samples common support for default configuration
* select common symbols by default when building with zephyr port
* add symbols to select between PSK and certificate auth
* add the hello sample


port/zephyr/CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ endif()
 zephyr_include_directories(include)
 zephyr_include_directories(../../src/include)
 
+add_subdirectory_ifdef(CONFIG_GOLIOTH_SAMPLE_COMMON ../../examples/zephyr/common CMAKE_BINARY_DIR)
```
mniestroj, line 8: This creates `modules/golioth-firmware-sdk/CMAKE_BINARY_DIR` in build directory. Was it intended?

mniestroj: Second thing. Maybe let's replace `../../` (which is actually pointing at root directory, with `${ZEPHYR_GOLIOTH_FIRMWARE_SDK_MODULE_DIR}/` (to be verified if that is the symbol that can be used).

port/zephyr/Kconfig
```diff
@@ -6,6 +6,16 @@ menuconfig GOLIOTH_FIRMWARE_SDK
...
+	select MBEDTLS
+	select MBEDTLS_DTLS if MBEDTLS_BUILTIN
+	select NET_SOCKETS
+	select NET_UDP
+	select ZCBOR
+	imply DNS_RESOLVER if NET_NATIVE
+	imply NET_SOCKETS_SOCKOPT_TLS
+	imply NET_SOCKETS_ENABLE_DTLS
+	help
+	  Enable library for communication with Golioth cloud.
```
mniestroj, line 18: Maybe let's remove those symbols from `examples/zephyr/golioth_basics` and actually see how much this simplifies `prj.conf`.

port/zephyr/Kconfig
```diff
@@ -21,6 +21,42 @@ if GOLIOTH_FIRMWARE_SDK
...
+	  Select single ciphersuite to reduce data exchanged during (D)TLS handshake, due to reduced
+	  ciphersuite list in Client Hello message.
+
+	  Make sure that credentials (e.g. using Zephyr TLS credentials subsystem) are configured
+	  for each ciphersuite (e.g. PSKs for PSK-based ciphersuites or certificates for
+	  certificate-based ciphersuites) that is negotiated.
+
+	  If empty, then underlying TLS implementation (e.g. mbedTLS library) decides which
+	  ciphersuites to use. Relying on that is not recommended!
 rsource '../../src/Kconfig'
```
mniestroj, line 60: 1. Please add newline before `resource`.
2. Maybe let's replace `rsource ../../` with `source ${ZEPHYR_GOLIOTH_FIRMWARE_SDK_MODULE_DIR}`?

port/zephyr/Kconfig
```diff
@@ -74,4 +74,6 @@ config GOLIOTH_ZEPHYR_THREAD_STACK_SIZE
 	help
 	  Threads stacks sizes for the use in golioth_sys_thread_create().
 
+rsource "../../examples/zephyr/common/Kconfig"
+
```
mniestroj, line 78: `examples/zephyr/common/Kconfig` is already sourced by `examples/zephyr/golioth_basics/Kconfig`. Let's not source the same file twice.

examples/zephyr/hello/README.md
```diff
@@ -54,10 +54,11 @@ CONFIG_GOLIOTH_SAMPLE_HARDCODED_KEY_PATH="keys/device.key.der"
 This application has been built and tested with QEMU x86 (qemu_x86).
 
 On your Linux host computer, open a terminal window, locate the source
-code of this sample application (i.e., `samples/hello`) and type:
+code of this sample application (i.e., `examples/zephyr/hello`) and
+type:
 
 ```console
-$ west build -b qemu_x86 samples/hello
+$ west build -b qemu_x86 examples/zephyr/hello
```
mniestroj: This is a fixup of the commit that introduced hello sample. Let's squash them.

port/zephyr/CMakeLists.txt
szczys: > This creates modules/golioth-firmware-sdk/CMAKE_BINARY_DIR in build directory. Was it intended?

No, that's a mistake. I added this because of the following build error when I don't specify a binary_dir:

```
CMake Error at /home/mike/golioth-compile/golioth-firmware-sdk/zephyr/cmake/modules/extensions.cmake:1637 (add_subdirectory):                                                 
  add_subdirectory not given a binary directory but the given source                                                                                                          
  directory                                                                                                                                                                   
  "/home/mike/golioth-compile/golioth-firmware-sdk/modules/lib/golioth-firmware-sdk/examples/zephyr/common"                                                                   
  is not a subdirectory of                                                                                                                                                    
  "/home/mike/golioth-compile/golioth-firmware-sdk/modules/lib/golioth-firmware-sdk/port/zephyr".                                                                             
  When specifying an out-of-tree source a binary directory must be explicitly                                                                                                 
  specified.                                                                                                                                                                  
Call Stack (most recent call first):                                                                                                                                          
  /home/mike/golioth-compile/golioth-firmware-sdk/modules/lib/golioth-firmware-sdk/port/zephyr/CMakeLists.txt:8 (add_subdirectory_ifdef)   
```

Is it more appropriate to use the following?

```
add_subdirectory_ifdef(CONFIG_GOLIOTH_SAMPLE_COMMON
    ${ZEPHYR_GOLIOTH_FIRMWARE_SDK_MODULE_DIR}/examples/zephyr/common common)
```

Or is there a better Cmake command for conditionally including that directory in the build?

port/zephyr/Kconfig
szczys: I have done so by adding this commit:
https://github.com/golioth/golioth-firmware-sdk/pull/129/commits/b9048cbcef4ca8fed4dd91ac6b80a2681cffc552

I was trying to keep golioth_basics as a more verbose example. Combining the default symbols from this commit with the common/Kconfig.defconfig committed later in this PR makes for nearly-empty prj.conf files in the more simplified samples.

examples/zephyr/common/Kconfig
```diff
@@ -4,6 +4,110 @@
...
+menuconfig GOLIOTH_SAMPLE_COMMON
+	bool "Common helper code shared across multiple samples"
+	help
+	  Build and link common code that is shared across Golioth samples.
+
+if GOLIOTH_SAMPLE_COMMON
+
+rsource "Kconfig.defconfig"
+
+choice GOLIOTH_AUTH_METHOD
```
sam-golioth, line 16: I think this should be `GOLIOTH_SAMPLE_HARDCODED_CRED_TYPE` or something along those lines. The SDK supports both PSK and Cert auth at runtime, so these symbols only apply to the hardcoded credentials module.

examples/zephyr/common/Kconfig
```diff
@@ -4,6 +4,110 @@
...
+	select MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
+	help
+	  Select mbedTLS dependencies for certificate based authentication.
+
+	  This is a convenience option when using mbedTLS as a library to make
+	  (D)TLS connection with Golioth cloud.
+
+if GOLIOTH_AUTH_METHOD_PSK
+
+config GOLIOTH_SAMPLE_PSK_ID
```
sam-golioth, line 69: Should these also be guarded under `GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS`?

examples/zephyr/common/Kconfig
```diff
@@ -31,3 +135,30 @@ config GOLIOTH_SAMPLE_WIFI_PSK
 	  unsecure (open) AP.
 
 endif # GOLIOTH_SAMPLE_WIFI
+
+config GOLIOTH_SYSTEM_SETTINGS
```
sam-golioth, line 139: This should be something like `GOLIOTH_SAMPLE_SETTINGS` or `GOLIOTH_SAMPLE_PSK_SETTINGS`. There's a similar change pending in https://github.com/golioth/golioth-zephyr-sdk/pull/406.

examples/zephyr/common/hardcoded_credentials.c
sam-golioth, line 1: I wonder if it's cleaner to have two versions of this file (one for PKI and one for PSK) and compile them conditionally? Instead of having the preprocessor switches in this file. Just a thought, I don't have a strong preference.

examples/zephyr/golioth_basics/Kconfig
```diff
@@ -3,12 +3,12 @@
 
 mainmenu "Golioth application options"
 
-config GOLIOTH_PSK_ID
+config GOLIOTH_SAMPLE_PSK_ID
```
sam-golioth, line 6: This is duplicated in `examples/zephyr/common/Kconfig` now, right?

examples/zephyr/hello/README.md
```diff
@@ -0,0 +1,177 @@
...
+$ west build -t run
+```
+
+See [Networking with
+QEMU](https://docs.zephyrproject.org/3.3.0/connectivity/networking/qemu_setup.html)
+on how to setup networking on host and configure NAT/masquerading to
+access Internet.
+
+#### ESP32
```
sam-golioth, line 70: Should we remove this until we have esp32 support in the CP SDK?

examples/zephyr/common/Kconfig
```diff
@@ -4,6 +4,110 @@
...
+
+config GOLIOTH_AUTH_METHOD_CERT
+	bool "Certificate based only"
+	imply GOLIOTH_AUTH_CERT_MBEDTLS_DEPS
+	help
+	  Only certificate based authentication will be supported.
+
+endchoice
+
+config GOLIOTH_AUTH_PSK_MBEDTLS_DEPS
```
sam-golioth, line 33: I'm not sure it makes sense to have this in the samples directory. Do we need it at all? Did it move from somewhere else?

examples/zephyr/hello/scripts/generate-keys.py
sam-golioth, line 1: Let's leave this out for now, until we have certificate support in the Zephyr support. There's an updated version of this script up for review in the Zephyr SDK, and we'll likely move this out of the Zephyr specific examples directory.

port/zephyr/CMakeLists.txt
mniestroj:
```
add_subdirectory_ifdef(CONFIG_GOLIOTH_SAMPLE_COMMON
    ${ZEPHYR_GOLIOTH_FIRMWARE_SDK_MODULE_DIR}/examples/zephyr/common common)
```
seems fine. Or maybe `examples_zephyr_common` instead of `common` just to make sure there are no conflicts with other directories.

The reason you've seen an error is because you provided relative directory path. In that case CMake requires you to provide explicitly what is the build directory (instead of using build/binary directory name based on source directory name). This makes sense. Consider you have `add_subdirectory(../common ...)` and `add_subdirectory(../../common ...)`. In that case there would be a conflict and build objects from both directories would land in the same build/binary directory. So keep that in mind when providing explicitly what the build/binary directory should be.

port/zephyr/Kconfig
mniestroj: There should be really just a single way how to configure examples. Otherwise such inconsistency is confusing to people wanting to compare those samples (e.g. "what does it take to configure more services at once, just like in golioth_basics sample").

examples/zephyr/common/Kconfig
mniestroj: > The SDK supports both PSK and Cert auth at runtime, so these symbols only apply to the hardcoded credentials module.

It shouldn't support both at runtime, at least with "production-like" configuration. Having support for all ciphersuites, even if not used, doesn't make sense. mbedTLS (or TLS in general) is taking a lot of resources and I've reduced its configuration a lot of times in the past with downstream projects, which grew up over time in terms of functionality, but there was just a limited resource pool (both Flash and RAM).

examples/zephyr/common/Kconfig
szczys: It's currently [in the main Kconfig of the Golioth Zephyr SDK](https://github.com/golioth/golioth-zephyr-sdk/blob/20c97b9cc3f6bd875c26a645734680f78718b96f/net/golioth/Kconfig#L29-L44). We had discussed moving this to samples as the desire was to generally remove PSK support from the CP SDK as only certs should be used in production.

These symbols are [used to populate the client config](https://github.com/golioth/golioth-firmware-sdk/blob/2939c177fc9fa0795e9101acb47d5846b332701d/examples/zephyr/common/hardcoded_credentials.c#L12-L40) in the samples.

examples/zephyr/golioth_basics/Kconfig
szczys: My intent was to leave golioth_basics as a more verbose example. So the client config is not handled by the common helper but shown in the app_main.c using these symbols. The prj.conf contains the symbols that would otherwise be set in the `common/Kconfig.defconfig` file.

sam-golioth: I see. It's confusing to have 2 symbols with the same name defined in different places, even if they aren't included in the same build ever. Can we change the name to make them distinct?

examples/zephyr/common/Kconfig
sam-golioth: I agree, we'll want to make it a compile time choice later. But that's not happening in this PR, and if it was, it wouldn't make sense to have these symbols defined in the examples Kconfig file. We can move the symbol and update the naming when we switch to selecting the auth type and cipher suites at build time.

examples/zephyr/common/Kconfig
sam-golioth: These `*_DEPS` symbols (and the symbols they select/imply) aren't being used in the code. But authentication was working previously without these options defined, so I'm unclear what it has to do with the examples?

> the desire was to generally remove PSK support from the CP SDK as only certs should be used in production

I think I could have been clearer on this point. We will support PSK and Cert auth in the CP SDK, but we will not support hardcoded credentials (of either type) in the core SDK code (they're only useful for examples and the quickest of prototyping), and we will not be opinionated about how to provision credentials (meaning that things like the settings service won't live in the core SDK).

szczys: The `*_DEPS` symbols are used in the hello sample. Without it, symbols are missing and communication doesn't function:

```
➜ diff build/zephyr/.config broken.config 
64c64
< CONFIG_GOLIOTH_CIPHERSUITES="TLS_PSK_WITH_AES_128_GCM_SHA256"
---
> CONFIG_GOLIOTH_CIPHERSUITES="TLS_PSK_WITH_AES_128_CBC_SHA256"
112c112
< CONFIG_GOLIOTH_AUTH_PSK_MBEDTLS_DEPS=y
---
> # CONFIG_GOLIOTH_AUTH_PSK_MBEDTLS_DEPS is not set
270c270
< CONFIG_MBEDTLS_KEY_EXCHANGE_PSK_ENABLED=y
---
> # CONFIG_MBEDTLS_KEY_EXCHANGE_PSK_ENABLED is not set
273,274d272
< CONFIG_MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED=y
< CONFIG_MBEDTLS_PSK_MAX_LEN=32
304c302
< CONFIG_MBEDTLS_CIPHER_GCM_ENABLED=y
---
> # CONFIG_MBEDTLS_CIPHER_GCM_ENABLED is not set
```

```
*** Booting Zephyr OS build zephyr-v3.4.0-553-g40d224022608 ***                                                                                                               
[00:00:00.030,000] <wrn> net_sock_tls: No entropy device on the system, TLS communication is insecure!                                                                        
[00:00:00.030,000] <inf> net_config: Initializing network                                                                                                                     
[00:00:00.030,000] <inf> net_config: IPv4 address: 192.0.2.1                                                                                                                  
[00:00:00.030,000] <dbg> hello_zephyr: main: start hello sample                                                                                                               
[00:00:00.030,000] <inf> golioth_samples: Waiting for interface to be up                                                                                                      
[00:00:00.040,000] <inf> golioth_mbox: Mbox created, bufsize: 1100, num_items: 10, item_size: 100                                                                             
[00:00:00.120,000] <inf> golioth_coap_client: Start CoAP session with host: coaps://coap.golioth.io                                                                           
[00:00:00.120,000] <inf> golioth_coap_client: Session PSK-ID: devboard-one-id@ttgo-demo                                                                                       
[00:00:00.120,000] <inf> golioth_coap_client: Entering CoAP I/O loop                                                                                                          
[00:00:02.290,000] <wrn> golioth_coap_client: CoAP message retransmitted                                                                                                      
[00:00:06.620,000] <wrn> golioth_coap_client: CoAP message retransmitted
```

This configuration is what is [used in the Golioth Zephyr SDK](https://github.com/golioth/golioth-zephyr-sdk/blob/bbae6f8419ed52acd74784ed7e876a0f24e75f8b/net/golioth/Kconfig#L32-L78), do you think we should approach it differently?

port/zephyr/CMakeLists.txt
szczys: Great, thanks for the advice. I've changed this to:

```
add_subdirectory_ifdef(CONFIG_GOLIOTH_SAMPLE_COMMON
    ${ZEPHYR_GOLIOTH_FIRMWARE_SDK_MODULE_DIR}/examples/zephyr/common examples_zephyr_common)
```

port/zephyr/Kconfig
szczys: To me it makes more sense to have this available without needing to source it in the Kconfig file of every project that wants to use it. So I've taken the source directive out of the `examples/zephyr/golioth_basics/Kconfig` file.

examples/zephyr/common/Kconfig
sam-golioth: Ah I see, this was getting set in golioth_basics/prj.conf before. I think it would make sense for these options to live in `port/zephyr/Kconfig` then.

examples/zephyr/common/hardcoded_credentials.c
szczys: I looked at this but I find it cleaner to have one file that handles both cases. 

Currently CMake looks like this:
```
zephyr_library_sources_ifdef(CONFIG_GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS hardcoded_credentials.c)
```

If we go with two files, I think we'd need to do something like this:
```
zephyr_library_sources_ifdef((CONFIG_GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS AND
    CONFIG_GOLIOTH_AUTH_METHOD_CERT) hardcoded_credentials_pki.c)
zephyr_library_sources_ifdef((CONFIG_GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS AND
    CONFIG_GOLIOTH_AUTH_METHOD_PSK) hardcoded_credentials_psk.c)
```
or I suppose this:
```
if(CONFIG_GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS)
    zephyr_library_sources_ifdef(CONFIG_GOLIOTH_AUTH_METHOD_CERT hardcoded_credentials_pki.c)
    zephyr_library_sources_ifdef(CONFIG_GOLIOTH_AUTH_METHOD_PSK hardcoded_credentials_psk.c)
}
```
But in both cases the C file will be nearly the same, with the config struct being the only difference.

I find it easier to understand the `#ifdef` in the C file. But if you prefer it the other way, let me know and I'll make the change.

sam-golioth: No preference, just a thought. I'm happy to keep it as is!

examples/zephyr/common/Kconfig
szczys: Yes, they should. Thanks for catching!

examples/zephyr/common/Kconfig
szczys: Okay, these now live in the `port/zephyr/Kconfig` file.

examples/zephyr/common/Kconfig
szczys: I discussed this with @sam-golioth and have moved the `GOLIOTH_AUTH_METHOD` related symbols to `ports/zephyr/Kconfig`.

port/zephyr/Kconfig
szczys: I discussed with @sam-golioth and have left these symbols as is for now.

examples/zephyr/common/Kconfig
szczys: I've guarded these symbols and I set the default for the hardcoded symbol to `n` this way it's more clear in the prj.conf file what's going on:

```
# Application
CONFIG_GOLIOTH_SAMPLE_COMMON=y
CONFIG_GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS=y
```

examples/zephyr/common/Kconfig
szczys: Sounds good, I've changed them to `GOLIOTH_SAMPLE_SETTINGS`.

examples/zephyr/golioth_basics/Kconfig
szczys: I made the following changes to golioth_basic regarding this conversation:
* Removed Kconfig symbols from the project in favor of those in the common library
* Added `CONFIG_GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS=y` so those symbols are available
* Left the client config struct in the example as that is the bit that I think is important for users to understand from this example.

port/zephyr/Kconfig
```diff
@@ -6,12 +6,110 @@ menuconfig GOLIOTH_FIRMWARE_SDK
...
+	  Enable library for communication with Golioth cloud.
 
 if GOLIOTH_FIRMWARE_SDK
 
 menu "Golioth SDK Configuration"
 
-rsource '../../src/Kconfig'
+rsource '${ZEPHYR_GOLIOTH_FIRMWARE_SDK_MODULE_DIR}/src/Kconfig'
+
+config GOLIOTH_CIPHERSUITES
```
sam-golioth, line 26: I don't think is being used anywhere, let's remove it.

sam-golioth (approved):

examples/zephyr/common/Kconfig.defconfig
```diff
@@ -0,0 +1,101 @@
...
+	default y if NETWORKING
+
+config MAIN_STACK_SIZE
+	default 4096
+
+config MINIMAL_LIBC_MALLOC_ARENA_SIZE
+	default 2048 if MINIMAL_LIBC
+
+config LOG_PROCESS_THREAD_STACK_SIZE
+	default 2048 if LOG_BACKEND_GOLIOTH
```
sam-golioth, line 79: There is no Golioth log backend yet, so we don't need this. Are there other settings in this file that go beyond what we need for these samples?

examples/zephyr/hello/prj.conf
```diff
@@ -0,0 +1,11 @@
...
+CONFIG_GOLIOTH_SAMPLE_COMMON=y
+CONFIG_GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS=y
+
+# Golioth Firmware SDK with all dependencies
+CONFIG_GOLIOTH_FIRMWARE_SDK=y
+
+CONFIG_FLASH=y
+CONFIG_FLASH_MAP=y
+CONFIG_STREAM_FLASH=y
+CONFIG_IMG_MANAGER=y
```
sam-golioth, line 11: Why do we need these flash and image symbles for a "hello" sample?

examples/zephyr/golioth_basics/prj.conf
```diff
@@ -12,17 +16,6 @@ CONFIG_NET_SHELL=y
...
-CONFIG_NET_UDP=y
-
-# Optional for some targets
-CONFIG_DNS_RESOLVER=y
-CONFIG_NET_SOCKETS_SOCKOPT_TLS=y
-CONFIG_NET_SOCKETS_ENABLE_DTLS=y
-
 # TLS configuration
 CONFIG_MBEDTLS_ENABLE_HEAP=y
 CONFIG_MBEDTLS_HEAP_SIZE=10240
```
sam-golioth, line 21: Some of these TLS options are enabled by `GOLIOTH_AUTH_PSK_MBEDTLS_DEPS` in `port/zephyr/Kconfig` so we don't need to set them here.

port/zephyr/Kconfig
szczys: I think we might need this. When I remove it I get this:
```
warning: the int symbol GOLIOTH_ZEPHYR_THREAD_STACK_SIZE (defined at /home/mike/golioth-compile/golioth-firmware-sdk/modules/lib/golioth-firmware-sdk/port/zephyr/Kconfig:121) has a non-int default GOLIOTH_COAP_THREAD_STACK_SIZE (undefined)
```
And I see that's in src/Kconfig:
```
➜ grep GOLIOTH_COAP_THREAD_STACK_SIZE ../../../src/Kconfig -A4
config GOLIOTH_COAP_THREAD_STACK_SIZE
    int "Golioth CoAP thread stack size"
    default 6144
    help
        Thread stack size of the Golioth CoAP thread, in bytes.
```

examples/zephyr/hello/prj.conf
szczys: It seems like the DFU portions of the SDK are built automatically and I don't see a way to turn them off (the Zephyr SDK has a [CONFIG_GOLIOTH_FW](https://github.com/golioth/golioth-zephyr-sdk/blob/bbae6f8419ed52acd74784ed7e876a0f24e75f8b/net/golioth/Kconfig#L139-L144) symbol for this).

These symbols are necessary to complete the build. But I don't think they should be part of the default config. I've left them here to be removed later when building DFU components is configurable.

sam-golioth: Works for me! Can you file a ticket to make DFU configurable?

port/zephyr/Kconfig
sam-golioth: Hmm, it doesn't seem like removing `GOLIOTH_CIPHERSUITES` should cause that particular error.

examples/zephyr/common/Kconfig.defconfig
szczys: Ah, good point. I've removed everything form defconfig that wasn't directly used in the current state of golioth_basics.

port/zephyr/Kconfig
szczys: @sam-golioth I was confused by what you were commenting on. I thought you were asking to remove the `rsource` line.

Sorry that I merged before reading this comment. I can make a new PR to remove this block.

sam-golioth: No worries! We're an agile team 😄

szczys: Ah, looking into this I remember why I added it. It's a value used in the twister testing:
https://github.com/golioth/golioth-firmware-sdk/blob/8a0825b3f32259287689a6f71b5ee7ad63ea939e/examples/zephyr/hello/sample.yaml#L40-L48

Do want me to remove the GOLIOTH_CIPHERSUITES along with that test?

PR #139, opened 2023-07-25
Zephyr log backend

sam-golioth:
~Depends on #140~


src/golioth_coap_client.c
```diff
@@ -830,6 +831,7 @@ static golioth_status_t coap_io_loop_once(
     if (!client->session_connected) {
         // Transitioned from not connected to connected
         GLTH_LOGI(TAG, "Golioth CoAP client connected");
+        golioth_sys_client_connected(client);
```
sam-golioth, line 834: The Zephyr log backend (which lives in the `port` directory) needs the `client` pointer in order to call `golioth_log_*` functions. I don't particularly like this pattern, but the others that I considered felt either less clean or more involved, so I picked this as the "least bad" option, and I'm open to feedback/suggestions on another way to get that pointer into the log module. My biggest concern is that I don't think the platform port should have to be aware of the Golioth system at all. The other options I considered were:
- Changing the log module so it doesn't need a client pointer passed in on every call - this is a bigger architectural change, and I like that the client context is a function argument - it makes the code easier to unit test (less global state).
- Having the platform ports register a callback - this doesn't address the fundamental question of whether the platform port should even know about the golioth service and is more complicated.

szczys (commented):
I tested this and reviewed the code and it LGTM. However, I defer to @mniestroj for the approval.

port/zephyr/golioth_sys_zephyr.c
```diff
@@ -193,6 +194,10 @@ void golioth_sys_thread_destroy(golioth_sys_thread_t gthread) {
...
 
-void golioth_sys_client_connected(void* client) {}
+void golioth_sys_client_connected(void* client) {
+    log_backend_golioth_enable(client);
+}
 
-void golioth_sys_client_disconnected(void* client) {}
+void golioth_sys_client_disconnected(void* client) {
+    log_backend_golioth_disable(client);
+}
```
mniestroj: Since those functions are implemented by platform-dependent code (Linux, Zephyr, ...) maybe we should put a `_platform_` in the function name?

port/zephyr/golioth_log_zephyr.c
```diff
@@ -0,0 +1,119 @@
...
+    lvl_to_log_func[log_msg_get_level(log)](
+        ctx->client, module, ctx->print_ctx.msg, GOLIOTH_WAIT_FOREVER);
+}
+
+static void init(const struct log_backend* const backend) {
+    log_backend_deactivate(backend);
+}
+
+static void panic(struct log_backend const* const backend) {
+    ((struct golioth_log_ctx*)backend->cb->ctx)->panic_mode = true;
```
mniestroj, line 89: Maybe let's unwrap it by adding one more variable? Personally I don't like (the readability of) such "cast and assign" statements.

mniestroj (commented):
Just minor comments.

One thing that I wanted to emphasize is that this implementation differs from the one in Zephyr SDK:
- CP SDK: just sends logs strings with formatted metadata (like function name, module name, timestamps, etc.) as part of string
- Zephyr SDK: it is more like "dictionary-style" logging, where metadata is separate from log string

port/zephyr/golioth_sys_zephyr.c
sam-golioth: We've been using `_sys_` for all of the platform dependent code so far, e.g. `golioth_sys_sem_create()`. I considered splitting out the more OS related code (semaphores, timers, threads, etc.) from the rest of it (right now just time, but would include these functions as well). That would also make it clearer for ports that use the same OS, like how ESP-IDF and Modus both use FreeRTOS. What do you think?

sam-golioth:
> Just minor comments.
> 
> One thing that I wanted to emphasize is that this implementation differs from the one in Zephyr SDK:
> 
> * CP SDK: just sends logs strings with formatted metadata (like function name, module name, timestamps, etc.) as part of string
> * Zephyr SDK: it is more like "dictionary-style" logging, where metadata is separate from log string

We can look at adding more metadata back later if we find that there is customer demand for it. For now, this keeps the implementation common among the different ports, and avoids a lot of string processing that was happening in the Zephyr port. I'm not convinced that we want all that happening by default for all customer applications - I prefer to keep it light weight.

port/zephyr/golioth_sys_zephyr.c
mniestroj: Okay, let's keep `_sys_` in the name. Makes sense!

mniestroj (approved):

PR #146, opened 2023-08-10
examples: zephyr: lightdb/get: remove credentials from prj.conf

mniestroj:
Remove PSK-ID/PSK credentials that were used during testing.


szczys (approved):
Apologies, thanks for catching this!

PR #150, opened 2023-08-14
linux: source PSK and PSK ID from environment

### LINKED ISSUE #149: Source PSK ID / PSK from Environment in Golioth Basics Linux Example
### Describe the feature

It can be helpful to spin up multiple `golioth_basics` examples when running from a local machine, each authenticating as a different device. However, PSK ID and PSK are currently sourced from `credentials.inc` at compile time, meaning that the example has to be compiled for each device connecting. It would me more convenient to allow for PSK ID and PSK to be sourced from environment variables at runtime. It would also absolve the user of needing to persist credentials in `credentials.inc` in their local repository.

Example of what a patch for updating to this behavior could look like:
```
🥽 (golioth-firmware-sdk) git diff ./examples/linux/golioth_basics/main.c
diff --git a/examples/linux/golioth_basics/main.c b/examples/linux/golioth_basics/main.c
index cd19b98..eb8e86a 100644
--- a/examples/linux/golioth_basics/main.c
+++ b/examples/linux/golioth_basics/main.c
@@ -2,31 +2,22 @@
 #include "golioth_basics.h"
 #include <string.h>
 #include <assert.h>
+#include <stdlib.h>
 
 #define TAG "main"
 
-// Include user's Golioth credentials.
-//
-// This is a required file, so if it doesn't exist, you will
-// need to create it and add the following:
-//
-//    #define GOLIOTH_PSK_ID "device@project"
-//    #define GOLIOTH_PSK "secret"
-//
-// The Golioth credentials can be found at:
-//
-//    console.golioth.io -> Devices -> (yourdevice) -> Credentials
-#include "credentials.inc"
-
 int main(void) {
+
+    char* golioth_psk_id = getenv("GOLIOTH_PSK_ID");
+    char* golioth_psk = getenv("GOLIOTH_PSK");
     golioth_client_config_t config = {
             .credentials = {
                     .auth_type = GOLIOTH_TLS_AUTH_TYPE_PSK,
                     .psk = {
-                            .psk_id = GOLIOTH_PSK_ID,
-                            .psk_id_len = strlen(GOLIOTH_PSK_ID),
-                            .psk = GOLIOTH_PSK,
-                            .psk_len = strlen(GOLIOTH_PSK),
+                            .psk_id = golioth_psk_id,
+                            .psk_id_len = strlen(golioth_psk_id),
+                            .psk = golioth_psk,
+                            .psk_len = strlen(golioth_psk),
                     }}};
 
     golioth_client_t client = golioth_client_create(&config);
```

I'm happy to make this update if folks are aligned 👍🏻 

### Existing feature reference

_No response_

hasheddan:
Updates the Linux Golioth Basics example to source PSK and PSK ID from environment variables so that recompiling is not necessary to provide new credentials.

Fixes #149

### DIFF
--- a/examples/linux/golioth_basics/main.c
+++ b/examples/linux/golioth_basics/main.c
@@ -2,31 +2,31 @@
 #include "golioth_basics.h"
 #include <string.h>
 #include <assert.h>
+#include <stdlib.h>
+#include <stdio.h>
 
 #define TAG "main"
 
-// Include user's Golioth credentials.
-//
-// This is a required file, so if it doesn't exist, you will
-// need to create it and add the following:
-//
-//    #define GOLIOTH_PSK_ID "device@project"
-//    #define GOLIOTH_PSK "secret"
-//
-// The Golioth credentials can be found at:
-//
-//    console.golioth.io -> Devices -> (yourdevice) -> Credentials
-#include "credentials.inc"
-
 int main(void) {
+    char* golioth_psk_id = getenv("GOLIOTH_PSK_ID");
+    if ((!golioth_psk_id) || strlen(golioth_psk_id) <= 0) {
+        fprintf(stderr, "PSK ID is not specified.\n");
+        return 1;
+    }
+    char* golioth_psk = getenv("GOLIOTH_PSK");
+    if ((!golioth_psk) || strlen(golioth_psk) <= 0) {
+        fprintf(stderr, "PSK is not specified.\n");
+        return 1;
+    }
+
     golioth_client_config_t config = {
             .credentials = {
                     .auth_type = GOLIOTH_TLS_AUTH_TYPE_PSK,
                     .psk = {
-                            .psk_id = GOLIOTH_PSK_ID,
-                            .psk_id_len = strlen(GOLIOTH_PSK_ID),
-                            .psk = GOLIOTH_PSK,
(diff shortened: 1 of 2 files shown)


examples/linux/golioth_basics/main.c
```diff
@@ -2,31 +2,22 @@
...
-//    #define GOLIOTH_PSK "secret"
-//
-// The Golioth credentials can be found at:
-//
-//    console.golioth.io -> Devices -> (yourdevice) -> Credentials
-#include "credentials.inc"
-
 int main(void) {
+    char* golioth_psk_id = getenv("GOLIOTH_PSK_ID");
+    char* golioth_psk = getenv("GOLIOTH_PSK");
```
sam-golioth, line 11: Can we add an error message if either of these don't exist in the environment?

sam-golioth (changes requested):

examples/linux/golioth_basics/main.c
hasheddan: Good call, will do!

hasheddan: @sam-golioth updated, let me know if there are any issues with error message style 👍🏻

sam-golioth (approved):

PR #155, opened 2023-08-18
zephyr: add rpc sample

szczys:
* add RPC sample
* set editorconfig to 72 char line width for Markdown files


examples/zephyr/rpc/src/main.c
```diff
@@ -0,0 +1,85 @@
...
+LOG_MODULE_REGISTER(rpc_sample, LOG_LEVEL_DBG);
+
+#include "golioth.h"
+#include <samples/common/hardcoded_credentials.h>
+#include <string.h>
+#include <zephyr/kernel.h>
+
+#include <samples/common/net_connect.h>
+
+golioth_client_t client;
```
mniestroj, line 17: That is not needed, right? Could you define it in `main()`?

szczys: Yes, I moved that as you recommended.

I also put all the includes together, without the space, in alphabetical order.

mniestroj (approved):

PR #156, opened 2023-08-22
golioth_rpc_status_t: use GOLIOTH_ namespace

szczys:
Add `GOLIOTH_` to all RPC status names so that the format matches: `GOLIOTH_RPC_OK`


szczys:
The RPC sample in #155 is related to this PR. That sample code needs to be updated when this PR is merged.

sam-golioth (approved):

PR #170, opened 2023-09-07
examples: zephyr: common: do not wait for dormant WiFi interface

sam-golioth:
DORMANT state in case of WiFi means that interface is ready for WiFi mgmt commands, such as "WiFi connect". This means that such interface will not get into UP state automatically and it makes no sense to wait for that event.

Above behavior was introduced some time ago in Zephyr upstream, however none of the WiFi drivers were updated yet to follow this behavior (remaining in DORMANT state before connecting to Access Point). Since the first driver (ESP-AT) was converted in Zephyr mainline, add support for such behavior.

Originally made by @mniestroj in Zephyr SDK.


hasheddan (approved):
LGTM! Thanks @sam-golioth (and @mniestroj!)!

PR #174, opened 2023-09-11
Add mimxrt1024 support

sam-golioth:
Adds support for the NXP MIMXRT1024-EVK board.

This also fixes a bug in the mailbox module. The mbox module had some of its logic wrapped in calls to `assert()`. When assertions are turned off, that logic was compiled out, and the mbox module stopped working. `assert()` should always only be used for testing values, and should never be used to wrap function calls, assignments, or other statements with side effects.


examples/zephyr/settings/boards/mimxrt1060_evkb.conf
```diff
@@ -0,0 +1,3 @@
+CONFIG_GOLIOTH_SAMPLE_DHCP_BIND=y
+CONFIG_NET_L2_ETHERNET=y
+CONFIG_NET_DHCPV4=y
```
mniestroj, line 3: Let's add `CONFIG_BUILD_OUTPUT_HEX=y` as in other samples.

mniestroj (commented):
Just a minor nit about `settings` example overlay.

Seems that we are missing lots of `sample.yaml` files, which means we are not build testing them in CI. Seems we should fix that at (doesn't need to be in this PR).

mniestroj (approved):

PR #178, opened 2023-09-13
Add zephyr beta readme

szczys:
* Document how to use the Zephyr beta on the Firmware SDK
* Add basic config/flashing info to zephyr golioth_basics (usage and console output will be added in the future)

### DIFF
--- a/examples/zephyr/README.md
+++ b/examples/zephyr/README.md
@@ -1,34 +1,110 @@
-# Golioth Zephyr Examples
+# Golioth Firmware SDK: Zephyr Support (beta)
 
-This directory contains a set of examples intended to build and run with the latest release of
-Zephyr.
+Zephyr support in the Golioth Firmware SDK is in active development and
+currently designated as "beta". This SDK supports all the features of
+the Golioth platform, but is not yet as optimized as the Golioth Zephyr
+SDK. We will update this page when this port is ready for general
+availability. For devices going to production in 2023, please use the
+[Golioth Zephyr SDK](https://github.com/golioth/golioth-zephyr-sdk).
 
-### Install Zephyr SDK
+We expect to reach general availability of Zephyr support in this SDK in
+early 2024. If you are building a Zephyr based device with plans to go
+to production in 2024 or later, you may want to consider beginning
+development in this SDK to reduce the burden of transitioning later. The
+SDK contains a set of examples designed to run using Zephyr RTOS.
 
-Follow [Zephyr Getting Started Guide](https://docs.zephyrproject.org/latest/develop/getting_started/).
+API documentation: <https://firmware-sdk-docs.golioth.io/>
 
-### Setup west project
+## Requirements:
 
-``` sh
-cd ~/golioth-workspace
+* Install the latest version of the [Zephyr
+  SDK](https://github.com/zephyrproject-rtos/sdk-ng/releases/latest).
+* Install [Python](https://www.python.org/downloads/)
+* Install [the West
+  tool](https://docs.zephyrproject.org/latest/develop/west/install.html)
+
+
+Most platforms are already supported with mainline [Zephyr
+RTOS](https://www.zephyrproject.org/). This repository can be added to
+any Zephyr based project as new West module. However, for making things
+simple, this repository can also serve as a West manifest repo.
+
+## Using the Golioth Firmware SDK as a West manifest repository
+
+### Using with mainline Zephyr
+
+Execute this command to download this repository together with all
+dependencies:
+
+```console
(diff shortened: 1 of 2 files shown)


examples/zephyr/README.md
```diff
@@ -1,34 +1,118 @@
-# Golioth Zephyr Examples
+# Golioth Firmware SDK: Zephyr Support (beta)
 
-This directory contains a set of examples intended to build and run with the latest release of
-Zephyr.
+Zephyr support in the Golioth Firmware SDK is in active development and
+currently designated as "beta". We will update this page when this port
```
sam-golioth, line 4: Adding more information on what "beta" means
```suggestion
currently designated as "beta". This SDK supports all the features of the Golioth platform, but is not yet as optimized as the Golioth Zephyr SDK. We will update this page when this port
```

examples/zephyr/README.md
```diff
@@ -1,34 +1,118 @@
...
 
-### Install Zephyr SDK
+The SDK contains a set of examples designed to run using Zephyr
+RTOS.
 
-Follow [Zephyr Getting Started Guide](https://docs.zephyrproject.org/latest/develop/getting_started/).
+API documentation: <https://firmware-sdk-docs.golioth.io/>
 
-### Setup west project
+## Install Golioth Firmware SDK with Zephyr RTOS Support
```
sam-golioth, line 13: I think it might be clearer to re-order these headings/sections, so that using this repo as the manifest repo (for either Zephyr or NCS) are grouped together, and adding the SDK to an existing west manifest is separated out (because it works for both NCS and Zephyr, it doesn't make sense to only have it under Zephyr). Here's what I'm thinking

## Using the Golioth Firmware SDK as a West manifest repository

### Using with mainline Zephyr

### Using with Nordic's nRF Connect SDK

## Adding the Golioth Firmware SDK to an existing Zephy-based West project

examples/zephyr/README.md
```diff
@@ -1,34 +1,118 @@
...
+* Install [the West
+  tool](https://docs.zephyrproject.org/latest/develop/west/install.html)
+
+
+Most platforms are already supported with mainline [Zephyr
+RTOS](https://www.zephyrproject.org/). This repository can be added to
+any Zephyr based project as new
+[west](https://docs.zephyrproject.org/3.4.0/develop/west/index.html)
+module. However, for making things simple, this repository can also
+serve as
```
sam-golioth, line 29:
```suggestion
serve as a
```

examples/zephyr/README.md
```diff
@@ -1,34 +1,118 @@
...
+Follow [Zephyr Getting
+Started](https://docs.zephyrproject.org/3.4.0/develop/getting_started/index.html)
+for details on how to setup Zephyr based projects.
 
-``` sh
-west build -b qemu_x86 modules/lib/golioth-firmware-sdk/examples/zephyr/golioth_basics
-west build -t run
+## Using Golioth with Nordic's nRF Connect SDK
+
+Platforms like [nRF9160
```
sam-golioth, line 73: I don't know that we need this sentence, but if we want an introduction here I think making it part agnostic makes sense (or at least refer to the part and not the board). Maybe something like:

Nordic maintains a fork of the Zephyr project with additional proprietary functionality for their silicon, such as cellular connectivity.

sam-golioth:
@dylanswartz This PR updates our READMEs and adds language about the beta and which SDK Zephyr developers should choose. Can you take a look?

examples/zephyr/README.md
dylanswartz: I like this

dylanswartz (commented):
I'll leave it to y'all to approve the PR but I like your wording updates @sam-golioth.

examples/zephyr/golioth_basics/README.md
```diff
@@ -4,3 +4,146 @@ This example will connect to Golioth and demonstrate Logging, Over-the-Air (OTA)
...
+## Requirements
+
+* Golioth credentials
+* Network connectivity
+
+## Building and Running
+
+### Authentication specific configuration
+
+#### PSK based auth
```
sam-golioth, line 17: Depending on who merges first, we may need to update this to include information about runtime PSK settings.

examples/zephyr/README.md
szczys: I think think you're right that we don't need this and have removed it in the latest push.

examples/zephyr/golioth_basics/README.md
szczys: This sample is set up to show how to use the config struct so it directly passes the kconfig symbols in main.c. We probably need to update the readme, the main.c, or both to match the approach taken in this example.

sam-golioth (approved):

PR #181, opened 2023-09-14
Docs fixes

sam-golioth:
We were previously running the doxygen workflow on the target branch, which isn't helpful. As a result a few bugs made it through.

- Test doxygen on PR branch
- Remove broken links
- Use Markdown formatted codeblocks instead of RST


sam-golioth:
Note that the `Deploy Doxygen to Firebase Hosting on PR / deploy_doxygen_dev (pull_request_target)` check will fail until this PR is merged. It is being triggered by the workflow file on the main branch and running against that branch, which is broken. This PR fixes it.

szczys (approved):
LGTM

PR #185, opened 2023-09-15
Initialize and update submodules in subshell

hasheddan:
Updates the Zephyr examples README.md to run the firmware SDK git submodule initialization and update in a subshell so that you remain in the top-level directory after running the command.


szczys:
I think the directory change is desireable. You want to be in the `modules/lib/golioth-firmware-sdk` directory to start a build.

hasheddan (commented):
@szczys fair enough 👍🏻 I was thinking in the context of building an app that depended on the firmware SDK, but you make a good point that this in the examples directory so presumably folks would be building one of the examples

PR #186, opened 2023-09-15
examples: zephyr: remove sample firmware version dependency on mcuboot

hasheddan:
Removes the dependency on BOOTLOADER_MCUBOOT for GOLIOTH_SAMPLE_FW_VERSION as it breaks DFU samples that do not use mcuboot.

Before this change I was seeing:
```
/home/hasheddan/code/github.com/golioth/golioth-firmware-sdk/examples/zephyr/dfu/src/main.c:18:39: error: 'CONFIG_GOLIOTH_SAMPLE_FW_VERSION' undeclared here (not in a function); did you mean 'CONFIG_GOLIOTH_SAMPLE_COMMON'?
   18 | static const char* _current_version = CONFIG_GOLIOTH_SAMPLE_FW_VERSION;
      |                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                       CONFIG_GOLIOTH_SAMPLE_COMMON
[148/396] Building C object zephyr/subsys/net/ip/CMakeFiles/subsys__net__ip.dir/net_context.c.obj
ninja: build stopped: subcommand failed.
FATAL ERROR: command exited with status 1: /usr/bin/cmake --build /home/hasheddan/code/github.com/golioth/golioth-firmware-sdk/examples/zephyr/dfu/build
```

After it compiled successfully.

### DIFF
--- a/examples/zephyr/common/Kconfig
+++ b/examples/zephyr/common/Kconfig
@@ -67,7 +67,6 @@ config GOLIOTH_SAMPLE_DHCP_BIND
 config GOLIOTH_SAMPLE_FW_VERSION
 	string "DFU Firmware Version"
 	default "0.0.0+0"
-	depends on BOOTLOADER_MCUBOOT
 	help
           Firmware version used by Golioth DFU sample to assign MCUBOOT_IMGTOOL_SIGN_VERSION (if
           using Zephyr) or MCUBOOT_IMAGE_VERSION (if using NCS)


szczys:
Interesting, I thought we needed `--sysbuild` for all non-NCS boards in order to build the bootloader and sign the images.

I just tried this and signed images are not preset. When I try the update it fails:
```
[00:02:38.179,000] <inf> fw_block_processor: Downloading block index 332 (333/334)
[00:02:38.264,000] <inf> fw_block_processor: Downloading block index 333 (334/334)
[00:02:38.348,000] <inf> golioth_fw_update: Download took 29803 ms
[00:02:38.349,000] <inf> fw_block_processor: Block Latency Stats:
[00:02:38.349,000] <inf> fw_block_processor:    Min: 42 ms
[00:02:38.349,000] <inf> fw_block_processor:    Ave: %.3f ms
[00:02:38.349,000] <inf> fw_block_processor:    Max: 256 ms
[00:02:38.349,000] <inf> fw_block_processor: Total bytes written: 341808
[00:02:38.349,000] <inf> golioth_fw_update: State = Downloaded
[00:02:38.437,000] <inf> golioth_fw_update: State = Updating
[00:02:38.525,000] <err> golioth_fw_update: Firmware change boot image failed
[00:02:38.525,000] <inf> golioth_fw_update: Waiting to receive OTA manifest
```

hasheddan (commented):
@szczys ah that makes sense! It looks like we just error out if `CONFIG_BOOTLOADER_MCUBOOT` is not enabled:
https://github.com/golioth/golioth-firmware-sdk/blob/9f19f5a145c1812db6ccdf6588a357aef7fe93fa/port/zephyr/golioth_fw_zephyr.c#L210

I see that we are setting it in the sysbuild conf here: https://github.com/golioth/golioth-firmware-sdk/blob/9f19f5a145c1812db6ccdf6588a357aef7fe93fa/examples/zephyr/dfu/sysbuild.conf#L1

If I'm not mistaken, you can build the mcuboot and application images separately (i.e. without `sysbuild`) and [flash them both on the device](https://docs.zephyrproject.org/latest/services/device_mgmt/dfu.html#mcuboot). In that case we would still want `CONFIG_BOOTLOADER_MCUBOOT` enabled, so I'm curious as to whether it makes sense to set it in the `prj.conf` as well, but remove the change I have made here.

szczys:
I much prefer the `--sysbuild` approach for two reasons:

1. The bootloader is automatically built, then automatically flashed when running `west flash`
2. Binaries are automatically signed

We used to build, sign, and flash separately and its quite a process. This really feels like the happy path. I just currently didn't work on this board when I tested.

I think it's better if the build fails if not using sysbuild. Since the binaries won't have been signed, and it may not be clear why the build worked, but the OTA update fails due to invalid binaries.

mniestroj:
> I think it's better if the build fails if not using sysbuild. Since the binaries won't have been signed, and it may not be clear why the build worked, but the OTA update fails due to invalid binaries.

That is one point of view. For me as SDK developer it makes sense to build without sysbuild and debug single application instead of "mcuboot+application" bundle. Maybe we should add a warning generated at cmake level when sysbuild is not enabled. But still, I want to be able to build example without sysbuild, mainly for development / troubleshooting purposes.

examples/zephyr/common/Kconfig
```diff
@@ -67,7 +67,6 @@ config GOLIOTH_SAMPLE_DHCP_BIND
 config GOLIOTH_SAMPLE_FW_VERSION
 	string "DFU Firmware Version"
 	default "0.0.0+0"
-	depends on BOOTLOADER_MCUBOOT
```
mniestroj, line 70: Please add `examples: zephyr: ` in commit title for better navigation across commits.

mniestroj (commented):
LGTM, just a minor nit.

examples/zephyr/common/Kconfig
hasheddan: @mniestroj done!

PR #189, opened 2023-09-19
ci: use previous commit for linting on main

hasheddan:
Updates lint workflow to use commit prior to push when merging to main. Previously the step would fail because $GITHUB_BASE_REF is empty.

See https://docs.github.com/en/webhooks/webhook-events-and-payloads#push:~:text=or%20null%20Required-,before,-string%20Required for more information.

Fixes golioth/firmware-issue-tracker#248

PR #147, opened 2023-08-10
support NCS and add nRF9160 DK configuration files

mniestroj:

PR #190, opened 2023-09-20
ci: increase fetch depth for linting

hasheddan:
Sets the fetch depth for linting to 10, which serves as a reasonable upper bound for the number of commits in a single PR. If the default fetch-depth of 1 is used, we are unable to calculate files changed as part of a PR being merged.

Follow-up to https://github.com/golioth/golioth-firmware-sdk/pull/189

### DIFF
--- a/.github/workflows/lint_build_unit_test.yml
+++ b/.github/workflows/lint_build_unit_test.yml
@@ -13,6 +13,10 @@ jobs:
       uses: actions/checkout@v3
       with:
         submodules: 'recursive'
+        # Fetch depth must be greater than the number of commits included in the push in order to
+        # compare against commit prior to merge. 15 is chosen as a reasonable default for the upper
+        # bound of commits in a single PR.
+        fetch-depth: 15
     - name: Setup Python
       uses: actions/setup-python@v2
       with:


.github/workflows/lint_build_unit_test.yml
```diff
@@ -13,6 +13,10 @@ jobs:
       uses: actions/checkout@v3
       with:
         submodules: 'recursive'
+        # Fetch depth must be greater than the number of commits included in the push in order to
+        # compare against commit prior to merge. 10 is chosen as a reasonable default for the upper
+        # bound of commits in a single PR.
+        fetch-depth: 10
```
sam-golioth, line 19: Let's make this 15. We do have some recent PRs with more than 10 commits, e.g. (https://github.com/golioth/golioth-firmware-sdk/pull/147)

sam-golioth (changes requested):

.github/workflows/lint_build_unit_test.yml
hasheddan: Sounds good!

sam-golioth (approved):

PR #191, opened 2023-09-20
examples: zephyr: cert provisioning: support esp32-devkitc-wrover

sam-golioth:
Add board configuration and overlay files for ESP32-DevKitC-WROVER, to match the other Zephyr examples.


hasheddan (approved):
Nice!

PR #192, opened 2023-09-21
Disable cid in ncs

sam-golioth:
NCS only supports CIDs for offloaded sockets, which we don't use (because of limitations on RSA key size in offloaded sockets I believe, but @mniestroj can correct me). Only select CID support when using with `MBEDTLS_BUILTIN` which we use as a proxy to mean "vanilla Zephyr version of mbedTLS".

Also, update to NCS 2.4.2 (from 2.4.0) to bring in latest fixes and improvements.


port/zephyr/Kconfig
```diff
@@ -11,7 +11,7 @@ menuconfig GOLIOTH_FIRMWARE_SDK
 	select REBOOT if BOOTLOADER_MCUBOOT
 	select REQUIRES_FULL_LIBC # libcoap
 	select MBEDTLS
-    select MBEDTLS_SSL_DTLS_CONNECTION_ID
+    select MBEDTLS_SSL_DTLS_CONNECTION_ID if MBEDTLS_BUILTIN
```
hasheddan, line 14: I think in practice `MBEDTLS_BUILTIN` is not set when we are using `nrf-sdk`, but it _could_ be. In that case I believe we could encounter issues -- are we okay with that for the time-being?

sam-golioth: Yeah, I'm okay with it. I think the chances of any of our current customers mucking with these settings is pretty low, and there's probably other things that would break too. We can revisit later, but I'm fine saying "using `MBEDTLS_BUILTIN` with NCS is unsupported" for now.

hasheddan (approved):

PR #193, opened 2023-09-21
examples: zephyr: stream: enable floats in printf

sam-golioth:
The LightDB Stream example sets floats directly. Internal to the SDK this is accomplished by using `snprintf()` to convert the float to a string. For this to work, floating point support in printf must be turned on in Kconfig.

Fixes golioth/firmware-issue-tracker#265


szczys (approved):
This is a fine short-term fix for this example.

In the long-run, we should find a way to either turn this on automatically or give the user notice at compile time that it needs to be enabled to use this function call.

PR #194, opened 2023-09-21
port: zephyr: enable fp support in printf for all libc

sam-golioth:
Turning on printf floating point support in newliib only solves the issue of sending flaots to lightdb stream for apps that use newlib. Select floating point support for each libc implementation. Also, move the selection to port/zephyr/Kconfig so it works for all apps not just the examples.


szczys (approved):
Yeah, this feels much cleaner!

PR #204, opened 2023-10-11
ci: parallelize twister builds

sam-golioth:
Creates a reusable workflow for running twister build tests on a single board, with either Zephyr or NCS. Uses this workflow to run build tests on each of our CVBs. Because these are run as separate steps in the caller workflow, they run in parallel on separate runners, and complete faster than if they were run sequentially.

Resolves golioth/firmware-issue-tracker#281

### DIFF
--- a/.github/workflows/twister-build.yml
+++ b/.github/workflows/twister-build.yml
@@ -1,0 +1,59 @@
+name: Twister build workflow
+
+on:
+  workflow_call:
+    inputs:
+      board:
+        required: true
+        type: string
+      manifest:
+        required: true
+        type: string
+      binary_blob:
+        required: false
+        type: string
+
+jobs:
+  twister_build:
+    container:
+      image: zephyrprojectrtos/ci:v0.26.4
+    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/twister-build.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  twister_build:
+    container:
+      image: zephyrprojectrtos/ci:v0.26.4
+    env:
+      ZEPHYR_SDK_INSTALL_DIR: /opt/toolchains/zephyr-sdk-0.16.1
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository and submodules
+      uses: actions/checkout@v3
```
beriberikix, line 25: v4?

beriberikix (approved):
LGTM! Left a comment to update an action since you were doing that in another PR.

.github/workflows/twister-build.yml
sam-golioth: #205 Updates this. I branched that change off of this one so that I could update all the action versions in one place.

PR #206, opened 2023-10-12
ci: use minimal docker image for build test

sam-golioth:
The official Zephyr CI Docker image is rather large, and includes many toolchains, modules, and other things that we don't need for our own builds. Using our own, smaller Docker image in CI can cut the time it takes to download and initialize the container from about 2-2.5 minutes down to about 30 seconds. This container does not come pre-installed with Zephyr's required python dependencies so this also adds a step in the CI workflow to install them, which is more robust anyway.

### DIFF
--- a/.github/workflows/twister-build.yml
+++ b/.github/workflows/twister-build.yml
@@ -15,8 +15,7 @@ on:
 
 jobs:
   twister_build:
-    container:
-      image: zephyrprojectrtos/ci:v0.26.5
+    container: golioth/golioth-zephyr-base:0.16.3-SDK-v0
     env:
       ZEPHYR_SDK_INSTALL_DIR: /opt/toolchains/zephyr-sdk-0.16.3
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


beriberikix:
Any idea why it failed?

beriberikix:
Ah, might be related to https://github.com/golioth/golioth-firmware-sdk/actions/runs/6499180729/job/17652019275#step:5:7. I've seen `request` issues in the past.

beriberikix:
I figured out the issue with the build (zephyrproject-rtos/zephyr/issues/63948,) and now can [successfully build](https://github.com/beriberikix/reference-design-template/actions/runs/6524677743) again.

Here's what I changed in the base image:
* [Moved the SDK](https://github.com/beriberikix/golioth-docker-base/blob/main/Dockerfile.zephyr#L20) to align what's in this workflow
* Added a [global override](https://github.com/beriberikix/golioth-docker-base/blob/main/Dockerfile.zephyr#L47) for Python / Ubuntu [warnings](https://peps.python.org/pep-0668/)
  * We'll need to decide if this is okay or move to either venv, a 3P [PPA](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa) or build from source (WHY PYTHON, WHYYYY)

I verified this with my test [RD template repo](https://github.com/beriberikix/reference-design-template). There's a few things I'm doing [there](https://github.com/beriberikix/reference-design-template/blob/main/.github/workflows/build_zephyr.yml). I think the only thing that needs to be added now is the [Setup West workspace](https://github.com/beriberikix/reference-design-template/blob/main/.github/workflows/build_zephyr.yml#L34-L41) step or similar. This was probably already done in the massive `CI` image.

mniestroj (approved):

beriberikix (approved):
Happy to see this adopted! We can switch to the Golioth-hosted one after that gets merged as well.

PR #208, opened 2023-10-16
zephyr: update to zephyr 3.5.0

sam-golioth:
Update to Zephyr 3.5.0

ESP32 crashes when using the stdlib implementations of `rand()` and `srand()`, but we shouldn't be using those anyway. The Zephyr random subsystem is tied into hardware support for HW RNGS and HW entropy, and uses a better PRNG than the stdlib implementation. So this PR adds support to the port layer for platform provided random functions, and updates the Zephyr port to use Zephyr random APIs instead of stdlib.

Unrelated (probably) to the above change, Zephyr renamed `rand32.h` to `random.h`, but NCS hasn't picked up that change yet, so we need to add a workaround for that.


src/include/golioth_sys.h
```diff
@@ -82,6 +82,19 @@ void golioth_sys_thread_destroy(golioth_sys_thread_t thread);
 #define golioth_sys_free(ptr) free((ptr))
 #endif
 
+/*--------------------------------------------------
+ * Random
+ *------------------------------------------------*/
+
+// Can be overridden via golioth_{user,port}_config
+#ifndef golioth_sys_srand
+#define golioth_sys_srand(seed) malloc((seed))
```
mniestroj, line 91: Why malloc?

sam-golioth: 🤦‍♂️ 

It was a copy paste error

mniestroj (approved):

PR #210, opened 2023-10-17
Ci Improvements

sam-golioth:
A couple small improvements to speed up build times:

Twister defaults to using twice the number of CPU cores when running build-only tests. Ninja also uses the number of CPU cores for building, meaning we could have 2\*N\*N (8 in our case) threads compiling simultaneously, which leads to bottlenecks and slows down builds. Experimentally, limiting twister to 2 threads (from a default of 4) results in the fastest build times.

We only need a subset of the projects that are in the main Zephyr and NCS West manifests in order to build our samples and tests. By using a manifest that only downloads the projects that we actually need, we can speed up our CI build times.

### DIFF
--- a/.ci-west-ncs.yml
+++ b/.ci-west-ncs.yml
@@ -1,0 +1,20 @@
+manifest:
+  self:
+    path: modules/lib/golioth-firmware-sdk
+    import:
+      file: west-ncs.yml
+      name-allowlist:
+        - cmsis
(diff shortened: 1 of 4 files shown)


beriberikix:
LGTM! I think it also makes sense to have a separate CI manifest :+1:

beriberikix (approved):

PR #214, opened 2023-10-25
workflows: used more vebose esp32s3_devkitc

szczys:
Change from esp32s3 to esp32s3_devkitc when targeting self-hosted runners. This more accurately describes the board connected, rather than just the chip type.


szczys:
The HIL test that failed earlier today now completes successfully: https://github.com/golioth/golioth-firmware-sdk/actions/runs/6646121403/job/18059030920

sam-golioth (approved):
Appreciate the quick turnaround!

PR #219, opened 2023-10-30
port: zephyr: enable DTLS 1.2 Connection IDs with NCS

hasheddan:
Removes MBEDTLS_BUILTIN conditional on enabling DTLS 1.2 Connection IDs. Previously, the NCS version used did not depend a Zephyr revision that included CID support. However, NCS v2.5.0 updated to a more recent revision that allows us to use CIDs even though we are not offloading DTLS to the modem.


hasheddan (commented):
Verified that prior to this change CIDs are not being used:

image


And afterwards they are:
Screenshot from 2023-10-30 15-02-02

mniestroj:
> Verified that prior to this change CIDs are not being used:
> 
> image
> 
> And afterwards they are: Screenshot from 2023-10-30 15-02-02

Just wanted to point out that above screenshots show outgoing packet size **increase** from 87 to 103 bytes. That is 18.4% increase in data consumption, at least for that simple and very minimal communication scenario.

PR #222, opened 2023-10-30
hil: add pytest plugin for boards

sam-golioth:
Adds a pytest plugin (`pytest-hil`) that provides a pytest fixture for interacting with boards connected over serial. This provides an abstracted API that is both easier to interact with when writing tests and also provides a consistent interface for performing actions on boards that may not have the same serial console semantics.

The plugin is installed with pip and is automatically picked up by pytest because of metadata in `pyproject.toml`.

HIL tests can simply add a `board` argument and they will be passed a `Board` object that is already initialized with an open serial port, PSK credentials, and WiFi credentials (if relevant). The tests can interact with the board by calling methods on the `Board` object, such as `board.reset()`. The details of how the board is actually reset are defined inside the plugin. The plugin is structured to use inheritance to reuse as much code as possible. For example, the nRF52840DK class inherits from the ZephyrBoard class, which inherits from the Board class. The plugin also defines command line arguments that are used to specify the board type along with any information needed to instantiate the board (such as credentials and the serial port path). These arguments are passed to `pytest` by the GitHub Actions workflow.

Closes golioth/firmware-issue-tracker#290


mniestroj:
So the idea is to create and maintain "cross-platform" pytest fixtures, of what is provided as part of Zephyr Pytest integration (which works obviously only with Zephyr now)?

sam-golioth:
> So the idea is to create and maintain "cross-platform" pytest fixtures, of what is provided as part of Zephyr Pytest integration (which works obviously only with Zephyr now)?

It's similar to what's provided by Zephyr, but we're not looking for a 1 to 1 replacement. Our fixture will be tailored to our needs, and I expect will be much smaller. And yes, the goal is to make it more "cross-platform".

tests/hil/scripts/pytest-hil/board.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @property
+    @abstractmethod
+    def PROMPT(self):
+        pass
+
+    @property
+    @abstractmethod
+    def USES_WIFI(self):
+        pass
```
mniestroj, line 51: Do we have some plan or coding guidelines when to use `UPPER_CASE` methods? I don't want to be too strict with tooling that is not core of our product, but I didn't see such naming convention before in Python projects, I think.

Personally I found `flake8` and `pylint` to be used most commonly, but this is a very subjective point of view. Overall, most Python linters or code checker tools are standardized around PEP 8: https://peps.python.org/pep-0008/ anyway.

mniestroj (commented):
Looks good overall.

Just a comment about coding style (more specifically UPPER_CASE method name).

tests/hil/scripts/pytest-hil/board.py
sam-golioth: The idea here is that `USES_WIFI()` should be a class constant defined by the subclasses. Python doesn't have constants, but it does have [class properties](https://docs.python.org/3/library/functions.html?highlight=property#property). Because we only define a getter for this property (and no setter), and because it returns a constant value (subclasses return either `True` or `False), it acts as constant. Because it has the property decorator, you can access it with `board.USES_WIFI` (no parenthesis). So even though it is defined as a method,  it is used as a constant property, which is why I made it upper case. I don't know of any guidelines around this kind of thing, but to me it makes the most sense and indicates the intention in the clearest way.

mniestroj: And how about using class attribute, like that:
```
class Board:
    USES_WIFI = False


class WiFiBoard(Board):
    USES_WIFI = True


class CellularBoard(Board):
    pass


if __name__ == '__main__':
    board = Board()
    wifi = WiFiBoard()
    cellular = CellularBoard()

    print('board', board.USES_WIFI)
    print('wifi', wifi.USES_WIFI)
    print('cellular', cellular.USES_WIFI)
```
which resolves to:
```
board False
wifi True
cellular False
```

sam-golioth: Class attributes are mutable:

```
class Board:
    USES_WIFI = False


class WiFiBoard(Board):
    USES_WIFI = True


class CellularBoard(Board):
    pass


if __name__ == '__main__':
    board = Board()
    wifi = WiFiBoard()
    cellular = CellularBoard()

    print('board', board.USES_WIFI)
    print('wifi', wifi.USES_WIFI)
    wifi.USES_WIFI = False
    print('wifi', wifi.USES_WIFI)
    print('cellular', cellular.USES_WIFI)
```
gives:
```
board False
wifi True
wifi False
cellular False
```

Using properties makes it immutable:

```
class Board:
    @property
    def USES_WIFI(self):
        return False


class WiFiBoard(Board):

    @property
    def USES_WIFI(self):
        return True


class CellularBoard(Board):
    pass


if __name__ == '__main__':
    board = Board()
    wifi = WiFiBoard()
    cellular = CellularBoard()

    print('board', board.USES_WIFI)
    print('wifi', wifi.USES_WIFI)
    wifi.USES_WIFI = False
    print('wifi', wifi.USES_WIFI)
    print('cellular', cellular.USES_WIFI)
```

```
board False
wifi True
Traceback (most recent call last):
  File "/Users/samfriedman/golioth/test_cp_sdk_zephyr/modules/lib/golioth-firmware-sdk/test.py", line 25, in <module>
    wifi.USES_WIFI = False
    ^^^^^^^^^^^^^^
AttributeError: property 'USES_WIFI' of 'WiFiBoard' object has no setter
```

mniestroj: I am not fond of using `@property` as a workaround for immutability (that is the nature of Python, there is even no concept of private or protected methods, since those are just "protected" by convention and nothing else). But But I don't won't to block just because of that reason.

sam-golioth: I think even more than just being immutable, it provides a cleaner getter/setter interface for variables, which is generally preferred over accessing object variables directly in OOP. In the future, if the logic for whether or not a board uses WiFi is more complex (e.g. if there is a SOC with WiFi and Cell support), the functions can be extended.

mniestroj: Actually there are plenty of libraries, or even all libraries, that use some kind of attributes. Either on the class or instance level. The same reasoning for immutability and getter/setter interface could be used for them. I can't imagine that converting all of the code to use `@property` in order to achieve immutability would be welcome. For the same reason I don't think it is a good idea to introduce such pattern in this module.

Besides, this property is written in UPPER_CASE, which means that it should not be overridden, just by convention.

> In the future, if the logic for whether or not a board uses WiFi is more complex (e.g. if there is a SOC with WiFi and Cell support), the functions can be extended.

In that case we should make that property instance specific and use snake_case name pattern.

Right now implementation says one thing, while name of the method suggests something else. Reliance on convention is quite important in Python, at least that is my feeling, so I am all for following it instead of inventing our own.

sam-golioth: There are also plenty of libraries using the @property decorator - this is a built-in feature of Python, so saying that it is uncommon or novel to use doesn't make much sense.

There are many ways of accomplishing the same goal - even in Python - and it's ok if the code I write doesn't look exactly like the code that you write and vice versa.

PR #223, opened 2023-10-31
ci: catch ram overflow build failures

sam-golioth:
Twister treats overflows in the RAM during builds as "skipped" tests instead of errors by default. Enabling the `--overflow-as-errors` option ensures we can catch overflows in CI.

Fixes golioth/firmware-issue-tracker#296

### DIFF
--- a/.github/workflows/twister-build.yml
+++ b/.github/workflows/twister-build.yml
@@ -49,6 +49,7 @@ jobs:
     - name: Run twister
       run: >
         zephyr/scripts/twister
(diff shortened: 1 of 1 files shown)


sam-golioth:
The build test failure is expected (it was already failing - this PR fixes twister so we actually detect it).

The HIL test failure is due to environment contamination on the runner and is expected for now - it's unrelated to these changes.

mniestroj (approved):

PR #225, opened 2023-11-01
examples: zephyr: fix esp32 build error in cert provisioning sample

sam-golioth:
Reduce the size of the DRAM 1 heap to prevent RAM overflow when building the certificate provisioning sample for the ESP32-DevKitC-WROVER board.

Fixes golioth/firmware-issue-tracker#299


szczys (approved):
LGTM! I test this and it works.

PR #226, opened 2023-11-01
port: zephyr: disable cid by default

sam-golioth:
Whether enabling connection IDs is beneficial or harmful is dependent on the application, so do not enable them by default. Instead, add a new Kconfig option for using connection IDs.


hasheddan (approved):
LGTM 👍🏻

mniestroj (approved):
There is a typo in commit title (s/disbale/disable/),  otherwise LGTM.

PR #227, opened 2023-11-02
tests: add program method to pytest hil fixture

sam-golioth:
Add the ability to program DUTs through the pytest-hil fixture, so that device specific programming logic can be removed from the GitHub Actions workflow yaml.

Closes golioth/firmware-issue-tracker#304

### DIFF
--- a/tests/hil/scripts/pytest-hil/nordicboard.py
+++ b/tests/hil/scripts/pytest-hil/nordicboard.py
@@ -1,0 +1,14 @@
+from board import Board
+from pynrfjprog import HighLevel
+
+class NordicBoard(Board):
+    def program(self, fw_image):
+        with HighLevel.API() as api:
+            with HighLevel.DebugProbe(api, int(self.serial_number)) as probe:
+                program_options = HighLevel.ProgramOptions(
+                    erase_action=HighLevel.EraseAction.ERASE_ALL,
+                    reset=HighLevel.ResetAction.RESET_SYSTEM,
+                    verify=HighLevel.VerifyAction.VERIFY_READ,
+                )
+
+                probe.program(fw_image, program_options=program_options)
--- a/tests/hil/scripts/pytest-hil/plugin.py
+++ b/tests/hil/scripts/pytest-hil/plugin.py
@@ -12,6 +12,10 @@ def pytest_addoption(parser):
         "--credentials-file", action="store", type=str,
         help="YAML formatted settings file"
     )
+    parser.addoption("--fw-image", type=str,
+            help="Firmware binary to program to device")
+    parser.addoption("--serial-number", type=str,
+            help="Serial number to identify on-board debugger")
 
 @pytest.fixture(scope="session")
 def board_name(request):
@@ -29,9 +33,18 @@ def baud(request):
 def credentials_file(request):
     return request.config.getoption("--credentials-file")
 
+@pytest.fixture(scope="session")
+def fw_image(request):
+    return request.config.getoption("--fw-image")
+
+@pytest.fixture(scope="session")
+def serial_number(request):
+    return request.config.getoption("--serial-number")
+
+
 @pytest.fixture(scope="module")
-def board(board_name, port, baud, credentials_file):
+def board(board_name, port, baud, credentials_file, fw_image, serial_number):
     if board_name.lower() == "nrf52840dk":
-        return nRF52840DK(port, baud, credentials_file)
+        return nRF52840DK(port, baud, credentials_file, fw_image, serial_number)
     else:
         raise ValueError("Unknown board")
--- a/.github/workflows/hil_test.yml
+++ b/.github/workflows/hil_test.yml
@@ -64,15 +64,7 @@ jobs:
         with:
           name: zephyr.hex
           path: .
-      - name: Flash build
-        run: |
-          source $HOME/runner_env.sh
-          nrfjprog --recover -f NRF52 --snr $CI_NRF52840DK_SNR
-          nrfjprog --program zephyr.hex --sectoranduicrerase --verify -f NRF52 --snr $CI_NRF52840DK_SNR
-          nrfjprog --pinresetenable -f NRF52 --snr $CI_NRF52840DK_SNR
-          nrfjprog --pinreset -f NRF52 --snr $CI_NRF52840DK_SNR
-          sleep 3
       - name: Run test
         run: |
           source $HOME/runner_env.sh
-          pytest --rootdir . tests/hil/tests/connection --board nrf52840dk --port $CI_NRF52840DK_PORT --credentials-file $HOME/credentials_nrf52840dk.yml
+          pytest --rootdir . tests/hil/tests/connection --board nrf52840dk --port $CI_NRF52840DK_PORT --credentials-file $HOME/credentials_nrf52840dk.yml --fw-image zephyr.hex --serial-number $CI_NRF52840DK_SNR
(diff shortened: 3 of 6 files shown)


tests/hil/scripts/pytest-hil/plugin.py
```diff
@@ -29,9 +33,18 @@ def baud(request):
 def credentials_file(request):
     return request.config.getoption("--credentials-file")
 
+@pytest.fixture(scope="session")
+def hex_file(request):
+    return request.config.getoption("--hex-file")
+
+@pytest.fixture(scope="session")
+def snr(request):
+    return request.config.getoption("--snr")
```
mniestroj, line 42: Nit. I think we could be more specific and not abbreviate too much here. `snr` is `serial_number` in this case, so I would prefer the latter.

tests/hil/scripts/pytest-hil/nordicboard.py
```diff
@@ -0,0 +1,18 @@
...
+            with HighLevel.DebugProbe(api, int(self.snr)) as probe:
+                program_options = HighLevel.ProgramOptions(
+                    erase_action=HighLevel.EraseAction.ERASE_ALL,
+                    reset=HighLevel.ResetAction.RESET_SYSTEM,
+                    verify=HighLevel.VerifyAction.VERIFY_READ,
+                )
+
+                probe.program(hex_file, program_options=program_options)
+
+        time.sleep(3)
```
mniestroj, line 18: Why do we have 3s sleep here? Maybe let's put some comment about it, why it is needed. Obviously it should state that it is a workaround, but the additional important information is what problem does it mitigate.

tests/hil/scripts/pytest-hil/plugin.py
```diff
@@ -29,9 +33,18 @@ def baud(request):
 def credentials_file(request):
     return request.config.getoption("--credentials-file")
 
+@pytest.fixture(scope="session")
+def hex_file(request):
+    return request.config.getoption("--hex-file")
```
mniestroj, line 38: This is enough for now, but some targets will be more complex than that. Especially that we might enter the "multi-component" firmwares at some point.

This comment is somehow related to https://github.com/golioth/golioth-firmware-sdk/pull/222#issuecomment-1789323641. Note that `west flash` with `sysbuild` addresses this issue to some point and at least allows to flash multi-component firmwares to a single SoC. This is the case for example for Cortex-M33 which have TF-M and two components for each security domain. In case of nRF5340 there is another components for the second core, which I believe (not 100% sure) can be handled by `sysbuild` (and `west flash` under `twister`) as well.

Even if we try to implement something more lightweight, the reality might be different. At least looking at the "design goal" which is to be "more cross-platform" implicitly means (or at least suggests) that this will be more complex than what Zephyr has already.

For now I wanted to point out the issues that we might encounter and express my concerns. One thought that goes through my mind is to reuse what Zephyr provides already in terms of:
- Pytest fixtures: there is shell interaction already
- twister: wraps multiple samples / boards and runs Pytest tests automatically based on various input parameters
- west runners (the backend for flashing targets, with integration of sysbuild for multi-target flashing)

mniestroj (commented):
This PR still provides obvious improvements over what we had so far, so good to go. However I pointed out my point of view and some concerns before we try to move too far in that direction (implementing **everything** on our own). We can consider reusing parts of Zephyr infrastructure or wrapping most of them to save time and maintenance effort in the long term.

mniestroj (approved):

tests/hil/scripts/pytest-hil/plugin.py
```diff
@@ -12,6 +12,10 @@ def pytest_addoption(parser):
         "--credentials-file", action="store", type=str,
         help="YAML formatted settings file"
     )
+    parser.addoption("--hex-file", type=str,
+            help="Firmware binary to program to device")
```
mniestroj, line 16: `--hex-file` might be too specific. At least in context of Espressif targets, used binary file has its own format, which is `.bin`, but with lots of embedded metadata used by `esptool.py`. So we might want to change that `--hex-file` either now or in the future, before futher "generalizing" the implementation for other platforms.

tests/hil/scripts/pytest-hil/plugin.py
sam-golioth: I looked into using `west flash`, the biggest issue is that `west flash` is designed to be used within a Zephyr development environment. When we build the binary in the cloud, we setup a Zephyr development environment to do the build. But when we run the test on the self-hosted runner, we only need the python scripts in this repo. So we only checkout this repository (and submodules - but I don't think we even need to that), and we don't install `west` or its python dependencies, and we don't run `west update` or otherwise download the Zephyr repo. Doing so would add a lot of overhead to the test and I don't think it's justified right now given how simple the `program()` code is.

I agree that when we have to test `sysbuild` builds (either for DFU or multicore chips, etc.) we may find that the complexity increases to the point that installing `west` and downloading the Zephyr repo becomes justified. But I would prefer to wait until we get to that point, and can explore other options as well.

mniestroj: > But when we run the test on the self-hosted runner, we only need the python scripts in this repo.

True, but this boils down to "pulling/installing more dependencies and get batteries-included tool" vs "minimal implementation for very specific use cases with higher development/maintenance effort".

The story with self-hosted runners is a bit different than using GitHub hosted runners. With self-hosted runners we have cache for free, which is actually the SD Card in case of HIL runners. SD cards are cheap and it is actually hard to get anything <32GB nowadays. This should be enough for installed Zephyr SDK, cached west modules and parts of build directory (from hosted runners). In that case Zephyr SDK installation would not be a runtime overhead (in terms of time spent). Cloning could be cache either by modifying `west update` or`git clone` commands or relying on cached workspace (see below, when I reference GitLab Runners default behavior). So in the end we would still just need to install Python scripts from this repo as part of the pipeline, since other dependencies would be cached/shared with maybe just a few seconds of overhead (reading few GB of data doesn't come entirely for free, especially when running rootfs on na SD Card).

> Doing so would add a lot of overhead to the test and I don't think it's justified right now given how simple the program() code is.

Not sure how it is with GitHub Runners, but in case of GitLab Runners the whole working directory (source + build artifacts) are cached for the next run. In that case `west update` just downloads repo once and then updates when necessary, i.e. when references in west manifest have changed. Something similar should be possible with GitHub Runners as well, which would lower the "overhead" of cloning to minimum.

sam-golioth: It's possible to have saved state on the runners, but we should avoid it. That can lead to both differing behavior between individual runners (based on their individual saved state) and also means that one run can cause problems with subsequent runs (by contaminating the saved state). This is actually already a problem for us, the HIL tests are failing on `main` because of environment contamination from the `pytest_plugin_for_boards` branch (in this case, a pip package). Mike has a ticket to start using docker images on the containers to avoid this issue. We can seed the docker image with Zephyr and West, but that has overhead too.

As far as the higher development/maintenance effort, it's already taken me longer to investigate using `west flash` on a separate machine from the one that did the build than it did to write the `program()` method. It's 10 lines of python, and it's unlikely to change often. Whereas ensuring that Zephyr and West are up-to-date and that our unsupported usecase (flashing on a different machine than building) keeps working is likely a nontrivial amount of ongoing work. Right now we don't have a complicated use case, and this is the most straightforward path. If supporting things like sysbuild in the future becomes more complicated, we can evaluate other solutions then.

PR #232, opened 2023-11-10
use python virtual environment for HIL tests

szczys:
Use a virtual environment to ensure that python packages installed on the system don't interfere with those installed by the test.

### DIFF
--- a/.github/workflows/hil_test.yml
+++ b/.github/workflows/hil_test.yml
@@ -56,7 +56,10 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
       - name: Setup Python dependencies
-        run: pip install pytest
+        run: |
+            python -m venv .venv
+            source .venv/bin/activate
+            pip install pytest pyserial PyYAML
       - name: Download build
         uses: actions/download-artifact@v3
         with:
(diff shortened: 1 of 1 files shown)


sam-golioth (changes requested):
Using `venv` isn't necessarily bad, but it doesn't provide the same kind of isolation that Docker does, and it doesn't help us distribute non-python tools to the runners in a consistent way (e.g. JLink).

Dan and Jon spent some time debugging environment issues with GitHub and Docker, and I believe that a similar fix will allow us to Docker on the runners, which would remove the need for `venv`.

See: https://github.com/golioth/fw-oci-images/pull/14

mniestroj:
I think that using venv makes sense regardless of using Docker or not. pip package manager has recently enforced a policy (or at least I noticed it on Arch Linux an Debian 12) that prevents installing Python packages in system directories, which makes venv always required. Maybe used Ubuntu Docker image does not have that yet as requirement, but still, using venv won't hurt anyway. The idea behind this policy of using venv for pip installed packages is to split distribution Python packages from pip, since those are maintained in separate way and this was common source of conflicts.

So overall, I am lean towards merging this PR.

mniestroj (approved):

hasheddan (commented):
Chatted with @szczys about this yesterday -- given that we are using the dependencies in every run here, it makes sense to run in a container image where they are already installed. If we move that direction, we can start venv as part of entrypoint on container startup. Anyone using the image, such as this workflow, shouldn't need to do any additional setup prior to invoking the tests 👍🏻

PR #234, opened 2023-11-13
ci: don't cancel other hil tests when one fails

sam-golioth:
The default GitHub Actions behavior is to cancel all other jobs spawned from a matrix when one of the jobs fails. This means our HIL tests are only able to identify at most one failure. Specifying `fail-fast` as false allows all jobs to complete even if one fails, giving us a more complete picture of how many tests are passing or failing.

### DIFF
--- a/.github/workflows/hil_tests.yml
+++ b/.github/workflows/hil_tests.yml
@@ -12,6 +12,7 @@ on:
 jobs:
   hil_test_zephyr:
     strategy:
+      fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


mniestroj (approved):
Makes sense, we use similar option in Zephyr SDK for parallel twister builds: https://github.com/golioth/golioth-zephyr-sdk/blob/97b9fb005d63ba5f80c354c89cbbc33e2d35f773/.github/workflows/build_test.yml#L21

PR #239, opened 2023-11-15
Fix intermittent RT1024 hil failures

sam-golioth:
Increase the size of the serial receive ringbuffer for the shell, to prevent the buffer from filling up and causing characters to be dropped. This can happen during HIL tests because if the DUT is running a higher priority thread, then the host PC can write to the serial port faster than the DUT empties the ringbuffer. This isn't a problem for the samples because humans don't type that fast, so this change is only made for the HIL test app.

This PR also includes a commit to erase boards that use JLink before programming them, which didn't fix the issue, but is a good idea anyway to keep things consistent and prevent one test run from influencing subsequent runs.

Fixes golioth/firmware-issue-tracker#312


mniestroj (approved):
I experienced a similar problem with QEMU some time ago, when developing some tests that used shell as well. Behavior was the same.

Alternative workaround would be to add inter-character sleep when sending commands over shell. But I guess this would not be any better, maybe even worse in potential debugging.

PR #245, opened 2023-12-04
Use docker for hil tests

szczys:
* Use docker for all HIL testing on this repo
* Specifically target on self-hosted runner for the legacy esp32s3 test to workaround timeouts on some runners. This is being phased out and will soon be replaced with more granular tests

Changing to Docker HIL images was tested before opening this PR:
* https://github.com/golioth/golioth-firmware-sdk/actions/runs/7089709731
* https://github.com/golioth/golioth-firmware-sdk/actions/runs/7067511221
* https://github.com/golioth/golioth-firmware-sdk/actions/runs/7066696608

### DIFF
--- a/.github/workflows/hil_test_esp-idf.yml
+++ b/.github/workflows/hil_test_esp-idf.yml
@@ -40,6 +40,13 @@ jobs:
     needs: build
     runs-on: has_${{ inputs.hil_board }}
 
+    container:
+      image: golioth/golioth-hil-base:ff78585
+      volumes:
+        - /dev:/dev
+        - /home/golioth/credentials:/opt/credentials
+      options: --privileged
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/hil_test_esp-idf.yml
```diff
@@ -40,6 +40,13 @@ jobs:
     needs: build
     runs-on: has_${{ inputs.hil_board }}
 
+    container:
+      image: golioth/golioth-hil-base:latest
```
sam-golioth, line 44: Let's point to a specific version here rather than the latest. That ensures we can keep building on this commit if we make a breaking change to the image later.

.github/workflows/hil_test_esp-idf.yml
```diff
@@ -40,6 +40,13 @@ jobs:
...
+      volumes:
+        - /dev:/dev
+        - /home/golioth/credentials:/opt/credentials
+      options: --privileged
```
sam-golioth, line 48: Why is this required? And what permissions does it grant?

PR #248, opened 2023-12-11
port: pass structs by reference in golioth_sys_* apis

sam-golioth:
Resolves golioth/firmware-issue-tracker#219

### DIFF
--- a/include/golioth_sys.h
+++ b/include/golioth_sys.h
@@ -45,7 +45,7 @@ typedef struct {
     void* user_arg;
 } golioth_sys_timer_config_t;
 
-golioth_sys_timer_t golioth_sys_timer_create(golioth_sys_timer_config_t config);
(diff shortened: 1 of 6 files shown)


include/golioth_sys.h
```diff
@@ -45,7 +45,7 @@ typedef struct {
     void* user_arg;
 } golioth_sys_timer_config_t;
 
-golioth_sys_timer_t golioth_sys_timer_create(golioth_sys_timer_config_t config);
+golioth_sys_timer_t golioth_sys_timer_create(golioth_sys_timer_config_t *config);
```
mniestroj, line 48: How about adding `const` to this argument type? Since this is a config, it probably should be mutable only in context of application and not modified by the SDK itself.

sam-golioth: Good thought! I've updated

mniestroj (approved):

PR #250, opened 2023-12-11
zephyr: add test for hello sample

szczys:
Add a pytest-based Twister test to the Hello sample:

* Compile and flash the sample to each HIL board
* Monitor the board's serial connection for the fourth Hello message to be sent
* Query the device logs from Golioth and search them backwards for the first three Hello messages

This test ensures the following:
* By waiting until the fourth message has been sent, we have a built-in 5 second window for the previous Hello message to arrive on the server and be available via the Rest API
* When searching the logs we verify all of the first three Hello messages were received in the order they were sent.

### DIFF
--- a/examples/zephyr/hello/pytest/conftest.py
+++ b/examples/zephyr/hello/pytest/conftest.py
@@ -1,0 +1,36 @@
+import os
+import pytest
+
+def pytest_addoption(parser):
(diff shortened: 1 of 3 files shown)


sam-golioth (approved):
Awesome work @szczys! No notes :)

PR #251, opened 2023-12-11
Kconfig cleanup

sam-golioth:
- Use bool instead of int or binary choices
- Remove redundant _ENABLE suffix
- Group kconfig options by service, and only show options when the relevant service is enabled
- Disable services by default

Note that only RPC and Settings are conditionally included based on Kconfig (this is how it already was). OTA, LightDB State, and LightDB Stream options will be added in future PRs.

Resolves golioth/firmware-issue-tracker#337


szczys (approved):
LGTM. I'm excited to see these added to the SDK. Looking forward for one that can switch DFU subsystem as well!

PR #252, opened 2023-12-13
zephyr: add test for the logging sample

szczys:
Check that each type of log massage is received, verifying both message text and log level.

### DIFF
--- a/examples/zephyr/logging/pytest/conftest.py
+++ b/examples/zephyr/logging/pytest/conftest.py
@@ -1,0 +1,36 @@
+import os
+import pytest
+
+def pytest_addoption(parser):
+    parser.addoption("--api-key", type=str,
+                     help="Golioth API key")
+    parser.addoption("--device-name", type=str,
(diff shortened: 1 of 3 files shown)


szczys:
@sam-golioth I don't understand why the esp32_devkitc_wrover test is getting skipped in this PR. Any thoughts?

https://github.com/golioth/golioth-firmware-sdk/actions/runs/7198154736/job/19607342471

sam-golioth (approved):
Great work @szczys! Let's file a ticket for the failures. Regarding those - are they consistent or intermittent?

szczys:
> Great work @szczys! Let's file a ticket for the failures. Regarding those - are they consistent or intermittent?

It is consistent that logs will not be found, but the logs that get dropped change between runs and between devices. I have seen some devices successfully pass the test, but never all devices during a single test.

szczys:
The behavior we are encountering is tracked in this issue: https://github.com/golioth/firmware-issue-tracker/issues/218

PR #255, opened 2023-12-14
examples: esp-idf: remove magtag demo

sam-golioth:
The hardware is out of stock, it's not one of our CVBs, and this kind of all-encompassing demo isn't in-line with our normal examples.

Resolves golioth/firmware-issue-tracker#343


szczys (approved):
LGTM!

PR #256, opened 2023-12-15
src: split lightdb into state and stream

sam-golioth:
The monolith lightdb module is split into lightdb_state, lightdb_stream, and payload_utils.

Additionaly, the lightdb_*_internal() helper functions were removed. These functions only abstracted 1 or 2 lines of common code each, but took 8-10 arguments, meaning that removing the helper functions actually improves readability by reducing the total lines of code by about 100.

Resolves golioth/firmware-issue-tracker#342


include/golioth_payload_utils.h
```diff
@@ -0,0 +1,53 @@
...
+
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
+
+
+/// @defgroup golioth_lightdb_state golioth_lightdb_state
+/// Functions for interacting with Golioth LightDB State service.
+/// @{
```
szczys, line 15: Is this specific to LightDB State or is it is an abstraction useful with multiple services?

szczys (approved):
LGTM

include/golioth_payload_utils.h
sam-golioth: Right now it's only used with State, but the intention was to make it generic, I just forgot to update the comment. Will fix, thanks!

PR #257, opened 2023-12-15
lightdb_state: add kconfig option for enabling lightdb_state service

sam-golioth:
Resolves golioth/firmware-issue-tracker#339


szczys (approved):
Should this Kconfig symbol be used in the `CMakeLists.txt` to conditionally include the `golioth_lightdb_state.c` in the build?

szczys:
> Should this Kconfig symbol be used in the `CMakeLists.txt` to conditionally include the `golioth_lightdb_state.c` in the build?

Ah, I see this is done with `#if defined()`, nevermind :blush:

sam-golioth:
> > Should this Kconfig symbol be used in the `CMakeLists.txt` to conditionally include the `golioth_lightdb_state.c` in the build?
> 
> Ah, I see this is done with `#if defined()`, nevermind 😊

Yep. It's done this way in the source file so that it also works for ports that don't use Kconfig + CMake (i.e. Linux and ModusToolbox). I have it in the back of my head for a future ticket to conditionally compile source files based off the Kconfig symbols, which would save a couple of seconds maybe during build, but we would still have to leave the `#if defined()` in there for the other platforms.

PR #259, opened 2023-12-15
HIL: reset USB hub before running tests

szczys:
Update Docker images for Twister and HIL base to newest which include USB hub control. Call the script to power cycle the hub before running tests.

### DIFF
--- a/.github/workflows/hil_sample_zephyr.yml
+++ b/.github/workflows/hil_sample_zephyr.yml
@@ -74,7 +74,7 @@ jobs:
     runs-on: has_${{ inputs.hil_board }}
 
     container:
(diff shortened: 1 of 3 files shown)


.github/workflows/hil_sample_zephyr.yml
```diff
@@ -105,6 +105,8 @@ jobs:
           pip3 install -r zephyr/scripts/requirements-run-test.txt
 
           pip3 install git+https://github.com/golioth/python-golioth-tools@v0.2.0
+      - name: Power Cycle USB Hub
```
sam-golioth, line 108: I suspect we'll also want to power cycle in between individual twister tests, but this is a good start!

sam-golioth (approved):

PR #263, opened 2023-12-19
src: remove golioth_statistics module

sam-golioth:
The golioth_statistics module is used to track memory allocations and detect memory leaks. It is a brittle, manual tool that requires every allocation and free to be marked with a call into the module. This clutters the code, is prone to breaking (if somebody forgets to add a tracking call or adds it improperly), and can add overhead. There are better ways to detect memory leaks including:

- using sanitizers
- using memory tools like valgrind
- using static analysis
- wrapping malloc and free to implement custom tracking
- using a gdb plugin

Resolves golioth/firmware-issue-tracker#351


szczys (approved):
I don't have an opinion on the philosophy of using or not using this approach to monitor for memory leaks. But I reviewed the code changes and they LGTM.

PR #264, opened 2023-12-19
remove remote_shell module

sam-golioth:
The remote shell module is an unfinished experiment that is currently not on our roadmap to productionize. The existing implementation is built on top of existing services (rpc and stream), is only implemented for ESP-IDF, and is not integrated into the web console for customers to use. A productionized version of remote shell would use a dedicated service (i.e. a new endpoint), requiring a rewrite of most of the existing code, so there's not much benefit to keeping it. There are costs however: keeping this rough proof of concept around increases our maintenance burden and could confuse users who might reasonably expect that we have (or will soon have) a working remote shell implementation.

Resolves golioth/firmware-issue-tracker#353


szczys (approved):
LGTM

PR #266, opened 2023-12-20
Lightdb state example tests

sam-golioth:
Add Twister tests for each of the LightDB State samples. Also includes a little bit of cleanup of error handling in the lightdb/get sample.

Resolves golioth/firmware-issue-tracker#274


szczys (approved):
LGTM

PR #267, opened 2023-12-21
Ensure disabled services fail to link

sam-golioth:
There's no reason to provide a noop function for a disabled SDK service. Instead, we should not provide implementations for disabled functions at all, so that if a user attempts to call these functions the build will fail at link time.

Fixes golioth/firmware-issue-tracker#355


szczys (approved):
LGTM!

PR #268, opened 2023-12-24
treewide: remove golioth_ file prefix

sam-golioth:
This is a big changelog, but it's just moving and renaming files. Header files are moved into a `golioth` folder and most of them have had the `golioth_` prefix removed, so now files are included by e.g. `#include <golioth/rpc.h>` instead of `#include "golioth_rpc.h"`. Similar changes are made for source files to match.

Resolves golioth/firmware-issue-tracker#345


szczys (approved):
LGTM. I suggest you add a CHANGELOG update as part of this commit (or as a second commit).

The double-quote includes never felt quite right, this approach fits much better, nice work!

sam-golioth:
> LGTM. I suggest you add a CHANGELOG update as part of this commit (or as a second commit).

Good call! I need to get better at doing that regularly in general.

PR #269, opened 2023-12-26
treewide: remove golioth.h

sam-golioth:
The golioth.h header just includes every other header. This reduces readability by hiding what's actually being used and included, and slightly slows down builds by including headers that aren't needed. Instead, source files should explicitly include headers for the modules that they actually use.

Resolves golioth/firmware-issue-tracker#346


szczys (approved):
LGTM!

This is a nice improvement. The only slightly odd one is payload_utils.h which is not as intuitive as the rest.
* I believe the C code for this file is not guarded by a Kconfig symbol so it is always built. If we're leaving it like that, we should just automatically include the header in any services that use callbacks with payloads.
* If we want to require user to include the header file, I think we should also require they select it in Kconfig.

sam-golioth:
> * I believe the C code for this file is not guarded by a Kconfig symbol so it is always built. If we're leaving it like that, we should just automatically include the header in any services that use callbacks with payloads.

I don't agree with this. The payload_utils are intended to be used by client application code, and if they are going to use it, then they should include it explicitly, not indirectly through another header.

> * If we want to require user to include the header file, I think we should also require they select it in Kconfig.

I don't think there's much advantage to having a Kconfig option for utility functions like this. They don't have any dependencies other than the standard library, and it's about 35 lines of code.

PR #270, opened 2023-12-26
examples: pytest: use golioth pytest plugin

sam-golioth:
Using the golioth pytest plugin (introduced in https://github.com/golioth/python-golioth-tools/pull/7) reduces the amount of boilerplate code that needs to be copy and pasted
between test cases.

Resolves golioth/firmware-issue-tracker#357

### DIFF
--- a/.github/workflows/hil_sample_zephyr.yml
+++ b/.github/workflows/hil_sample_zephyr.yml
@@ -104,7 +104,7 @@ jobs:
           pip3 install -r zephyr/scripts/requirements-build-test.txt
           pip3 install -r zephyr/scripts/requirements-run-test.txt
 
-          pip3 install git+https://github.com/golioth/python-golioth-tools@v0.2.0
+          pip3 install git+https://github.com/golioth/python-golioth-tools@v0.3.0
       - name: Power Cycle USB Hub
         run: /opt/golioth-scripts/power-cycle-usb-hub.sh
       - name: Download build
--- a/examples/zephyr/hello/pytest/test_sample.py
+++ b/examples/zephyr/hello/pytest/test_sample.py
@@ -1,4 +1,3 @@
-from golioth import Client
 import logging
 import pytest
 import pprint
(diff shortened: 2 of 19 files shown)


.github/workflows/hil_sample_zephyr.yml
```diff
@@ -104,7 +104,7 @@ jobs:
           pip3 install -r zephyr/scripts/requirements-build-test.txt
           pip3 install -r zephyr/scripts/requirements-run-test.txt
 
-          pip3 install git+https://github.com/golioth/python-golioth-tools@v0.2.0
+          pip3 install git+https://github.com/golioth/python-golioth-tools@7fad3fe3216a15d42fc9a35f50c434b9b8eba062
```
sam-golioth, line 107: This ref should be changed to `v0.3.0` after https://github.com/golioth/python-golioth-tools/pull/7 is merged and the merge commit is tagged.

examples/zephyr/hello/pytest/test_sample.py
```diff
@@ -11,15 +10,7 @@
...
-async def test_hello(shell, api_key, device_name, credentials_file):
-
-    # Connect to Golioth and get device object
-
-    client = Client(api_url = "https://api.golioth.dev",
-                    api_key = api_key)
-    project = (await client.get_projects())[0]
-    device = await project.device_by_name(device_name)
-
+async def test_hello(shell, device, credentials_file):
```
szczys, line 13: I don't understand where `device` is coming from.

I expected to see a change in the workflow that passes this in but that's not the case.

sam-golioth: `device` is provided by the new pytest plugin that was added to `python-golioth-tools`. The change in the workflow is subtle - it's the update to the revision of the python package that we install as part of the workflow. This is one of those things that's both slick and frustrating about pytest plugins. Slick because all a testcase needs to do is add an argument to its function and presto, pytest provides that argument. Frustrating because all that magic makes it non-obvious what's going on under the hood.

mniestroj (approved):

PR #271, opened 2023-12-26
Conditionally include fw update

sam-golioth:
Only include the fw update service and its dependencies if the user selects it. This allows for smaller binaries for applications that do not use the fw update service.

Also update the Zephyr examples that don't use DFU to not include its required dependencies (flash, img_mgr, etc.).

Resolves golioth/firmware-issue-tracker#223


szczys (approved):
LGTM!

PR #272, opened 2023-12-26
examples: zephyr: rename dfu to fw_update

sam-golioth:
Firmware Update is what we call the service in the Golioth console, and matches what we call the service in the SDK source. It's also a more widely used term than dfu and so should be clearer to a wider range of users.

Resolves golioth/firmware-issue-tracker#360


szczys (approved):
LGTM

PR #274, opened 2023-12-27
lightdb_state: take content type for get/set

sam-golioth:
Change lightdb_set_json_\*() to lightdb_set_\*() and add a `content_type` argument so they can also accept CBOR formatted payloads.

Similar changes are done for lightdb_get operations that previously only worked with JSON.

The examples are updated, as well as the twister tests.

Resolves golioth/firmware-issue-tracker#347


src/ota.c
```diff
@@ -12,7 +12,7 @@
 #include "coap_client.h"
 #include <golioth/golioth_debug.h>
 #include "golioth_util.h"
-#include "zcbor_utils.h"
+#include <golioth/zcbor_utils.h>
```
mniestroj, line 15: What is the reason to expose that header publicly? (There is no reasoning in the commit message.)

sam-golioth: So that applications can use it. In this case, I'm using it in the lightdb/get example for CBOR.

mniestroj (approved):

PR #277, opened 2024-01-03
Remove unnecessary typedefs

sam-golioth:
We were using typedefs in 2 ways, both of which were problematic:

Using typedefs to create `void *` handles for the coap client. Typedef'ing a pointer prevents the `const` keyword from working correctly, and casting to `void *` prevents compile-time type checks. We want to hide the members of a struct, not the fact that the variable is a pointer to a struct, so we can use opaque structs instead.

Other times we would create a new type with typedef, but still expect the user to know the underlying type. E.g. by typedef'ing a struct and then requiring users to set or read members of that struct. In this case, using a typedef obscures information about the variable, but doesn't actually provide any encapsulation or abstraction. The end result is just code that's harder to read.

Resolves golioth/firmware-issue-tracker#362

### DIFF
--- a/include/golioth/client.h
+++ b/include/golioth/client.h
@@ -97,18 +97,18 @@ struct golioth_pki_credential {
     size_t private_key_len;
 };
 
-/// TLS Authentication Credentials
-typedef struct {
+/// TLS Authentication Credential
+struct golioth_credential {
     enum golioth_auth_type auth_type;
     union {
         struct golioth_psk_credential psk;
         struct golioth_pki_credential pki;
     };
-} golioth_tls_credentials_t;
+};
 
 /// Golioth client configuration, passed into golioth_client_create
 typedef struct {
-    golioth_tls_credentials_t credentials;
+    struct golioth_credential credentials;
 } golioth_client_config_t;
 
 /// Callback function type for client events
--- a/include/golioth/client.h
+++ b/include/golioth/client.h
@@ -50,7 +50,7 @@ struct golioth_response {
     ///      GOLIOTH_ERR_TIMEOUT (no response received from server)
     ///      GOLIOTH_OK (2.XX)
     ///      GOLIOTH_ERR_FAIL (anything other than 2.XX)
-    golioth_status_t status;
+    enum golioth_status status;
     /// the 2 in 2.XX
     uint8_t status_class;
     /// the 03 in 4.03
@@ -199,7 +199,7 @@ bool golioth_client_wait_for_connect(struct golioth_client* client, int timeout_
 ///
 /// @return GOLIOTH_OK Client started
 /// @return GOLIOTH_ERR_NULL Client handle invalid
-golioth_status_t golioth_client_start(struct golioth_client* client);
+enum golioth_status golioth_client_start(struct golioth_client* client);
 
 /// Stop the Golioth client
 ///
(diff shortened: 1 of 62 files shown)


include/golioth/golioth_debug.h
```diff
@@ -16,7 +16,7 @@
 #define LOG_TAG_DEFINE(tag) static __UNUSED const char* TAG = #tag
 #endif
 
-typedef void* golioth_client_t;
+struct golioth_client;;
```
szczys, line 19:
```suggestion
struct golioth_client;
```

It looks like there is an extra semicolon on his one.

include/golioth/client.h
```diff
@@ -57,13 +57,13 @@ struct golioth_response {
     uint8_t status_code;
 };
 
-/// TLS authentication type
-typedef enum {
+/// Aauthentication type
```
szczys, line 60:
```suggestion
/// Authentication type
```

include/golioth/client.h
```diff
@@ -65,19 +65,19 @@ enum golioth_auth_type {
     GOLIOTH_TLS_AUTH_TYPE_PKI,
 };
 
-/// Pre-Shared Key (PSK) credentials.
+/// Pre-Shared Key (PSK) credential.
 ///
 /// All memory is owned by user and must persist for the lifetime
 /// of the golioth client.
-typedef struct {
+struct golioth_psk_credential {
```
szczys, line 72: Here you have also changed the naming of the struct by dropping the `s`.

To me this is more confusing because there are multiple data in this struct. Is this change based on practices we're already using in the SDK? It might be good to mention this in the commit message.

include/golioth/client.h
```diff
@@ -79,11 +79,11 @@ struct golioth_psk_credential {
     size_t psk_len;
 };
 
-/// Public Key Infrastructure (PKI) credentials (aka "certificates").
+/// Public Key Infrastructure (PKI) credential (aka "certificate").
 ///
 /// All memory is owned by user and must persist for the lifetime
 /// of the golioth client.
-typedef struct {
+struct golioth_pki_credential {
```
szczys, line 86: Same comment as I left on the PSK credential. It would be to mention the reasoning for the name change (dropping the `s`) in the commit message.

include/golioth/client.h
```diff
@@ -97,18 +97,18 @@ struct golioth_pki_credential {
     size_t private_key_len;
 };
 
-/// TLS Authentication Credentials
-typedef struct {
+/// TLS Authentication Credential
+struct golioth_credential {
```
szczys, line 101: Same comment on mentioning the name change (dropped `s`) reasoning in the commit message.

szczys (approved):
I left comment with a few nits but LGTM.

I might be a good idea to update the CHANGELOG as part of this PR to track the breaking changes.

include/golioth/client.h
sam-golioth: There is only one credential in the struct: A PSK-ID/PSK pair. The same is true for the other credential structs - each one holds a single credential.

mniestroj (approved):

include/golioth/client.h
sam-golioth: Keeping it as singular, but I updated the commit message for this and the others.

PR #278, opened 2024-01-04
Remove circular references between services and coap_client

sam-golioth:
The coap_client module should not have any knowledge of the services built on top of it. Remove the settings and rpc structs from the coap_client and instead create init functions that creates the state structs for each, adds a pointer to the client, and returns the struct to the user.

Resolves golioth/firmware-issue-tracker#370

### DIFF
--- a/include/golioth/rpc.h
+++ b/include/golioth/rpc.h
@@ -9,11 +9,12 @@
 #include <zcbor_encode.h>
 #include <golioth/golioth_status.h>
 #include <golioth/client.h>
-#include <golioth/config.h>
 
 /// @defgroup golioth_rpc golioth_rpc
 /// Functions for interacting with the Golioth Remote Procedure Call service
 /// @{
+
+struct golioth_rpc;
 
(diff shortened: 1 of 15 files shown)


mniestroj (approved):

include/golioth/rpc.h
```diff
@@ -82,9 +83,17 @@ typedef enum golioth_rpc_status (*golioth_rpc_cb_fn)(
...
 ///
 /// @param client Golioth client handle
+///
+/// @return pointer to golioth rpc struct
+/// @return NULL - Error initializing RPC service
+struct golioth_rpc* golioth_rpc_init(struct golioth_client* client);
+
+/// Register an RPC method
+///
+/// @param grpc Golioth client handle
```
szczys, line 96: Does this doxy param ref need updating?

szczys (approved):

include/golioth/rpc.h
sam-golioth: Good catch!

PR #281, opened 2024-01-06
Detect and fix modem reset loop

sam-golioth:
The nrf91 modem has a built in reset loop restriction that activates when the device is reset 7 times in a row within 30 seconds of connecting to the network. This restriction prevents the modem from connecting to the network for 30 minutes. We hit this frequently during CI. This PR monitors for an event from the modem that indicates the reset loop is active and notifies the user when the event is received. It also adds a Kconfig option to automatically factory reset the modem when we enter the reset loop. This removes the reset loop restriction and lets the modem start working immediately, but is not recommended in production because a) it may just send you back into the reset loop and b) factory reset writes to flash and so has flash wear implications. Neither of these apply to us in CI because a) we know why the device is resetting so often (we're doing it on purpose) and b) we're writing to the program flash more often than we do a factory reset, so the MCU flash will wear out before the modem flash.

Fixes golioth/firmware-issue-tracker#317

### DIFF
--- a/examples/zephyr/common/Kconfig
+++ b/examples/zephyr/common/Kconfig
@@ -87,6 +87,16 @@ config GOLIOTH_SAMPLE_NRF91_LTE_MONITOR
 	help
 	  LTE Link Control events monitor for nRF91.
 
+config GOLIOTH_SAMPLE_NRF91_RESET_LOOP_OVERRIDE
+	bool "Automatically override modem reset loop restriction"
+	depends on GOLIOTH_SAMPLE_NRF91_LTE_MONITOR
+	help
+	  When selected, the device will automatically factory reset the modem
+	  when the modem reports that it has activated the reset loop restriction.
+	  This will remove the reset loop restriction.
+
+	  WARNING: It is not recommended to activate this setting in production.
+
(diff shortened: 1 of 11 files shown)


mniestroj (approved):

examples/zephyr/common/Kconfig
```diff
@@ -87,6 +87,16 @@ config GOLIOTH_SAMPLE_NRF91_LTE_MONITOR
...
+config GOLIOTH_SAMPLE_NRF91_RESET_LOOP_OVERRIDE
+	bool "Automatically override modem reset loop restriction"
+	depends on GOLIOTH_SAMPLE_NRF91_LTE_MONITOR
+	help
+		When selected, the device will automatically factory reset the modem
+		when the modem reports that it has activated the reset loop restriction.
+		This will remove the reset loop restriction.
+
+		WARNING: It is not recommended to activate this setting in production.
+
```
mniestroj, line 99:
```suggestion
	  When selected, the device will automatically factory reset the modem
	  when the modem reports that it has activated the reset loop restriction.
	  This will remove the reset loop restriction.

	  WARNING: It is not recommended to activate this setting in production.

```

PR #283, opened 2024-01-06
ci: use auto-generated boards for twister tests

sam-golioth:
Don't provide the `GOLIOTH_DEVICE_NAME` environment variable to the twister tests. This will cause the golioth pytest plugin to automatically generate a temporary device in the console to use for the duration of the test, after which it will be deleted. Using automatically generated devices reduces the amount of manual work needed to add a device into the HIL runners (we no longer need to manually create devices) and will make it easier to move to a different project or backend environment in the future.

This uses the functionality provided by golioth/python-golioth-tools#9 and golioth/python-golioth-tools#10

Resolves golioth/firmware-issue-tracker#326

### DIFF
--- a/.github/workflows/hil_sample_zephyr.yml
+++ b/.github/workflows/hil_sample_zephyr.yml
@@ -119,7 +119,6 @@ jobs:
         env:
           hil_board: ${{ inputs.hil_board }}
           GOLIOTH_API_KEY: ${{ secrets.DEV_CI_PROJECT_API_KEY }}
-          GOLIOTH_DEVICE_NAME: ${{ runner.name }}-${{ inputs.hil_board }}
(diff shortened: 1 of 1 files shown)


.github/workflows/hil_sample_zephyr.yml
```diff
@@ -104,7 +104,7 @@ jobs:
           pip3 install -r zephyr/scripts/requirements-build-test.txt
           pip3 install -r zephyr/scripts/requirements-run-test.txt
 
-          pip3 install git+https://github.com/golioth/python-golioth-tools@56f34269a1a25cb34049fe3a484f5a0801a5db78
+          pip3 install git+https://github.com/golioth/python-golioth-tools@40d7bddf0502c98df47292bd2fe4e05e7534b34e
```
sam-golioth, line 107: Update to a tag before merging

mniestroj (approved):

PR #287, opened 2024-01-11
workflows: use tag for legacy esp-idf test

szczys:
Use device tag when creating release so that it doesn't interfere with other OTA tests that may run at the same time.


szczys:
The critical test for this change is `Test on Hardware / build_for_hw_test (pull_request)` which has passed CI so this is ready for review.

sam-golioth (approved):

PR #290, opened 2024-01-15
Add fw_update sample test

szczys:
Test OTA by running the fw_update example on all available HIL board types.

* This targets prod so I'm basing this merge on the `test_against_prod` branch
* nRF9160 currently failing but that's actually the test catching an issue with our SDK

### DIFF
--- a/.github/workflows/hil_sample_zephyr.yml
+++ b/.github/workflows/hil_sample_zephyr.yml
@@ -59,6 +59,7 @@ jobs:
         zephyr/scripts/twister                                                \
               --platform ${{ inputs.west_board }}                             \
               -T modules/lib/golioth-firmware-sdk/examples/zephyr             \
+              --build-only                                                    \
               --package-artifacts test_artifacts_${{ inputs.west_board }}.tar
 
     - name: Save artifacts
--- a/.github/workflows/hil_sample_zephyr.yml
+++ b/.github/workflows/hil_sample_zephyr.yml
@@ -104,7 +104,7 @@ jobs:
           pip3 install -r zephyr/scripts/requirements-build-test.txt
           pip3 install -r zephyr/scripts/requirements-run-test.txt
 
-          pip3 install git+https://github.com/golioth/python-golioth-tools@v0.5.0
+          pip3 install git+https://github.com/golioth/python-golioth-tools@v0.5.1
       - name: Power Cycle USB Hub
         run: /opt/golioth-scripts/power-cycle-usb-hub.sh
       - name: Download build
--- a/.github/workflows/hil_sample_zephyr.yml
+++ b/.github/workflows/hil_sample_zephyr.yml
@@ -59,7 +59,6 @@ jobs:
         zephyr/scripts/twister                                                \
               --platform ${{ inputs.west_board }}                             \
               -T modules/lib/golioth-firmware-sdk/examples/zephyr             \
-              --prep-artifacts-for-testing                                    \
               --package-artifacts test_artifacts_${{ inputs.west_board }}.tar
 
     - name: Save artifacts
@@ -118,6 +117,7 @@ jobs:
         shell: bash
         env:
           hil_board: ${{ inputs.hil_board }}
+          west_board: ${{ inputs.west_board }}
           GOLIOTH_API_KEY: ${{ secrets.PROD_CI_PROJECT_API_KEY }}
           GOLIOTH_CREDENTIALS_FILE: /opt/credentials/credentials_${{ inputs.hil_board }}.yml
         run: |
--- a/examples/zephyr/fw_update/pytest/test_sample.py
+++ b/examples/zephyr/fw_update/pytest/test_sample.py
@@ -1,0 +1,104 @@
+from golioth import Client
+import logging
+import pytest
+import time
+import yaml
+
+LOGGER = logging.getLogger(__name__)
+
+UPDATE_VERSION = '255.8.9'
+UPDATE_PACKAGE = 'main'
+API_URL = "https://api.golioth.io"
+CI_PROJ_NAME = "firmware_ci"
+
+pytestmark = pytest.mark.anyio
+
+async def test_fw_update(shell, project, device, credentials_file, west_board):
+    # Read credentials
+
+    with open(credentials_file, 'r') as f:
+        credentials = yaml.safe_load(f)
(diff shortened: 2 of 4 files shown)


.github/workflows/hil_sample_zephyr.yml
```diff
@@ -104,7 +103,7 @@ jobs:
           pip3 install -r zephyr/scripts/requirements-build-test.txt
           pip3 install -r zephyr/scripts/requirements-run-test.txt
 
-          pip3 install git+https://github.com/golioth/python-golioth-tools@1afc52f6b76a0518fb080f1e055336499b66fd45
+          pip3 install git+https://github.com/golioth/python-golioth-tools@5a02dfe3697b6281dcf2e8ddd99986067e3482d0
```
szczys, line 106: We should cut a release of the python-golioth-tools and target it here.

examples/zephyr/fw_update/pytest/test_sample.py
```diff
@@ -0,0 +1,107 @@
...
+import time
+import yaml
+import datetime
+import os
+
+LOGGER = logging.getLogger(__name__)
+
+UPDATE_VERSION = '255.8.9'
+UPDATE_PACKAGE = 'main'
+API_URL = "https://api.golioth.io"
```
szczys, line 13: FYI this targets prod (so the workflow needs to also use the prod API key secret)

examples/zephyr/fw_update/pytest/test_sample.py
```diff
@@ -0,0 +1,107 @@
...
+
+    device_check = await project.device_by_name(device.name)
+
+    d_package = device_check.info['metadata']['update']['main']['package']
+    d_version = device_check.info['metadata']['update']['main']['version']
+
+    assert d_package == 'main', f'Expected firmware package "main" but got "%s"'.format(d_package)
+    assert d_version == '255.8.9', f'Expected firmware version "255.8.9" but got "%s"'.format(d_version)
+
+    # Clean Up
```
szczys, line 103: @sam-golioth is there a way to run this cleanup step if one of the asserts ends the test before we get here?

examples/zephyr/fw_update/pytest/test_sample.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # Monitor block download and watch for reboot after update
+
+    shell._device.readlines_until(regex=".*Downloading block index.", timeout=90.0)
+    LOGGER.info("Block download has begun!")
+
+    shell._device.readlines_until(regex=".*Rebooting into new image.", timeout=600.0)
+    LOGGER.info("Download complete, restarting to perform update.")
+
+    shell._device.readlines_until(regex=".*Current firmware version: main - 255.8.9.", timeout=90.0)
```
szczys, line 87: This is currently causing the nRF9160 to fail because the log message is never written (although the OTA update completes successfully). I'm tracking this issue here: https://github.com/golioth/firmware-issue-tracker/issues/390

examples/zephyr/fw_update/pytest/test_sample.py
```diff
@@ -0,0 +1,107 @@
...
+
+    time.sleep(2)
+
+    # Add blueprint to device
+
+    board_name = os.environ['west_board']
+    LOGGER.info(f"Board name: {board_name}")
+
+    client = Client(api_url = API_URL, api_key = os.environ['GOLIOTH_API_KEY'])
+    project = await client.project_by_name(CI_PROJ_NAME)
```
szczys: I just added a commit to test the alternative approach to requesting `project` from the fixture. This can be squashed once the CI is done running.

examples/zephyr/fw_update/pytest/test_sample.py
sam-golioth: Based on our conversation during 1:1 I think this comment is on the wrong line? It's referring to the "Rebooting into new image" message on line 84, correct?

examples/zephyr/fw_update/pytest/test_sample.py
```diff
@@ -0,0 +1,107 @@
...
+
+    shell._device.readlines_until(regex=".*Current firmware version: main - 255.8.9.", timeout=90.0)
+    LOGGER.info("Device reported expected update version")
+
+    shell._device.readlines_until(regex=".*Nothing to do.", timeout=30.0)
+    LOGGER.info("Device received manifest confirming successful upgrade.")
+
+    # Check cloud for reported firmware version
+
+    device_check = await project.device_by_name(device.name)
```
sam-golioth, line 95: I wonder if we should add a `device.refresh()` function or something.

examples/zephyr/fw_update/pytest/test_sample.py
sam-golioth: For posterity, in 1:1 we discussed using pytest fixtures to perform cleanup

sam-golioth (approved):

.github/workflows/hil_sample_zephyr.yml
```diff
@@ -104,7 +104,7 @@ jobs:
           pip3 install -r zephyr/scripts/requirements-build-test.txt
           pip3 install -r zephyr/scripts/requirements-run-test.txt
 
-          pip3 install git+https://github.com/golioth/python-golioth-tools@v0.5.0
+          pip3 install git+https://github.com/golioth/python-golioth-tools@4e02e8fb41fc79d3d3df2a54501076bfe7578f5c
```
szczys, line 107: This should be updated to `v0.5.1` once https://github.com/golioth/python-golioth-tools/pull/21 has merged.

szczys: This has merged and I have updated the tag in the workflow.

examples/zephyr/fw_update/pytest/test_sample.py
sam-golioth: It would mean that we can reuse variables instead of creating new ones. Keeping old, stale device objects around increases the likelihood that they will be used by accident. Having a `refresh()` method or similar will keep the code cleaner and make it harder to accidentally use stale data. Having it as part of the API is also a hint to users that a device object is a snapshot instead of a live view (and so needs to be "refreshed" to get the latest data).

Anyway, just a thought for the future, no action needed for this PR.

PR #291, opened 2024-01-17
Update coding style

sam-golioth:
This updates our coding standard and applies those changes across the codebase.

The first two commits fix our CI check for coding standard compliance, which had been detecting coding standard violations but still reported a pass to GitHub actions, and removes some wrapper scripts that were adding complexity without adding much value.

The third commit updates our coding standard and applies it across the tree, while the fourth commit makes some minor code updates to produce better looking results when formatted with clang-format. These commits are best reviewed together. The most important thing to review here are the changes to the `.clang-format` file to make sure you're okay with the standards it lays out. The [docs](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) have good explanations for each option, with examples. It would be good to give the rest of the changes to the tree a quick skim, but don't feel like you need to closely examine all 5000 line changes.

The fifth commit updates our brace style and applies it across the codebase. I kept this commit separate because I expect it to be the most controversial change. I switched us to using the Allman brace style, which keeps all open and close curly braces (`{` and `}`) on their own line. I believe that this is more readable because open and closing braces are always on the same indentation level, and the mostly empty line containing the brace provides a visual break between blocks of code. The common argument against this style is that it's not as compact as others, but we should be optimizing for readability - not compactness - and I don't believe that making the code more compact increases readability. In fact, in many cases I believe it reduces readability.

Resolves golioth/firmware-issue-tracker#385


.clang-format
```diff
@@ -18,6 +18,7 @@ ContinuationIndentWidth: 4
 IndentCaseLabels: true
 IndentWidth: 4
 IndentWrappedFunctionNames: false
+IndentExternBlock: NoIndent
```
sam-golioth, line 21: Note that this isn't actually being applied correctly right now because of a bug in clang-format that was fixed a few weeks ago. We need to wait for the next release to get the fix.
https://github.com/llvm/llvm-project/issues/36620
https://github.com/llvm/llvm-project/pull/75731

.clang-format
```diff
@@ -73,7 +74,7 @@ AlwaysBreakTemplateDeclarations: true
 BinPackArguments: false
 BinPackParameters: false
 BreakBeforeBinaryOperators: NonAssignment
-BreakBeforeBraces: Attach
+BreakBeforeBraces: Allman
```
szczys, line 77: I have some thoughts on this :smiley: 

* My personal preference is `BreakBeforeBraces: Stroustrup`.
    * We can both agree that breaking before functions makes a lot of sense
    * I like the look and feel of `if (condition) {` all on one line
    * With Allman I find the else conditions are very awkward:

        ```c
          }
          else
          {
        ```

    * These feel much more natural with Stroustrup:

        ```c
          }
          else {
        ```

* FWIW, [Zephyr is on Linux style](https://github.com/zephyrproject-rtos/zephyr/blob/18c23dec5f60836fc84a9cc93293769dff84e5c2/.clang-format#L30)

sam-golioth: > I like the look and feel of `if (condition) {` all on one line

Do you find that it helps you to read the code better? I find that having the open and closing braces in the same column makes it easier to visually line them up. And having a mostly empty line at the start and end of a block helps delineate them. For example in this code, it's easier to see at a glance where the conditional stops and the body starts in Allman (2nd) vs Stroustrup (1st):

```
if (is_this_function_true() &&
    get_some_value() == MY_CONSTANT &&
    check_value(foo)) {
    call_a_function(bar);
    baz = x + 5;
    printf("something something something");
}
```

```
if (is_this_function_true() &&
    get_some_value() == MY_CONSTANT &&
    check_value(foo))
{
    call_a_function(bar);
    baz = x + 5;
    printf("something something something");
}
```

> With Allman I find the else conditions are very awkward

I agree they're verbose, but they don't interfere with my ability to read and understand the code.

szczys: Yeah, that makes sense. I'm fine with going with the one we think is more readable.

szczys (approved):

mniestroj (approved):

PR #296, opened 2024-01-19
zephyr: CI: add Hello test using CID

szczys:
Add a Hardware in the Loop test that uses connection ID for the Hello sample.

### DIFF
--- a/examples/zephyr/hello/sample.yaml
+++ b/examples/zephyr/hello/sample.yaml
@@ -16,6 +16,21 @@ tests:
       mimxrt1024_evk
       nrf52840dk_nrf52840
       nrf9160dk_nrf9160_ns
+  sample.golioth.hello.cid:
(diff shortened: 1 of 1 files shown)


examples/zephyr/hello/sample.yaml
```diff
@@ -16,6 +16,21 @@ tests:
       mimxrt1024_evk
       nrf52840dk_nrf52840
       nrf9160dk_nrf9160_ns
+  sample.golioth.hello.cid:
```
sam-golioth, line 19: Let's also remove the `sample.goloth.hello.psk.cid` build test on line 42. Testing CID with either hardcoded or runtime PSK should be sufficient to uncover issues with CID specifically so I don't think we need both, and I'd rather not extend our CI build times.

sam-golioth (approved):

PR #301, opened 2024-01-25
stream: use golioth_content_type to set objects

szczys:
Remove the `golioth_stream_set_cbor_async/sync` and `golioth_stream_set_json_async/sync` functions and replace them with `golioth_stream_set_async/sync` function that requires a `golioth_content_type` as an argument.

This makes it possible to add more content types in the future without introducing breaking changes.


sam-golioth (approved):
Looks good! Just need to rebase/retarget

PR #304, opened 2024-01-25
Add migration guide

szczys:


docs/Migration_Guide_Zephyr.md
```diff
@@ -0,0 +1,587 @@
...
+Golioth repository to rule them all. From the API standpoint, the change
+is mostly naming and syntax. We'll walk you through how to make those
+updates in your code. Under the hood, this change means we can develop,
+test, and rollout new features and better performance to all of our
+supported platforms so you always have access to the best Golioth has to
+offer.
+
+The [Golioth Zephyr SDK](https://github.com/golioth/golioth-zephyr-sdk)
+will be deprecated later in 2024. This migration guide details the
+actions necessary to an existing Zephyr project to the Golioth Firmware
```
sam-golioth, line 17:
```suggestion
actions necessary to move an existing Zephyr project to the Golioth Firmware
```

docs/Migration_Guide_Zephyr.md
```diff
@@ -0,0 +1,587 @@
...
+- [Contact the Golioth Developer Relations
+  team](mailto:devrel@golioth.io) with migration questions.
+
+## Migration Overview
+
+1. [Update `west` manifest](#update-west-manifest)
+2. [Update Kconfig files](#update-kconfig-files)
+3. [Update header file includes](#update-header-file-includes)
+4. [Change Golioth API calls](#change-golioth-api-calls)
+5. [Special case: Changes for OTA Firmware
```
sam-golioth, line 36: These special case links don't seem to work in the "rich diff" preview in Github

docs/Migration_Guide_Zephyr.md
```diff
@@ -0,0 +1,587 @@
...
+```
+
+The naming scheme for these files has been updated so that it is easier
+to understand what includes do. This comes with the added benefit of
+making it easier to intuitively know what header file you need.
+
+### Sample header files
+
+Golioth example code uses a common sample library. This is not
+technically a part of the core SDK (it servers as a guide for developing
```
sam-golioth, line 181:
```suggestion
technically a part of the core SDK (it serves as a guide for developing
```

docs/Migration_Guide_Zephyr.md
```diff
@@ -0,0 +1,587 @@
...
+> and the [Golioth Firmware
+> SDK](https://github.com/golioth/golioth-firmware-sdk/tree/main/examples/zephyr)
+> that replaces it.
+
+### 1. Creating the Golioth client
+
+- The Golioth client now requires a `golioth_client_config` struct that
+  contains a credentials for authenticating with Golioth. This is used
+  for either certificate authentication or pre-shared key
+  authentication.
```
sam-golioth, line 210: My suggested wording might need work, but we should provide information about how to use the credential tag functionality that Marcin is adding to enable offloaded DTLS sockets
```suggestion
  for either certificate authentication, pre-shared key
  authentication, or credential tag authentication.
```

docs/Migration_Guide_Zephyr.md
```diff
@@ -0,0 +1,587 @@
...
+    ```
+
+- Pass a pointer to the configuration when creating the Golioth client:
+
+    ```c
+    golioth_client_t client = golioth_client_create(&client_config);
+    ```
+
+- The callback that runs when Golioth connects no longer needs to be
+  registered, and now passes a `golioth_client_event` into the function:
```
sam-golioth, line 245: The callback does still need to be registered, but it is registered through a function call instead of setting a pointer in the struct:
`golioth_client_register_event_callback(client, on_client_event, NULL);`

sam-golioth (changes requested):
Looks good! Just a couple content comments and a few typos.

docs/Migration_Guide_Zephyr.md
szczys: Ah, thanks for catching this. I've made an update to reference the new function call.

szczys (commented):
Thanks for the feedback, I made the following changes:

* Fix the table of contents in the two commits they were originally added
* Fix the on_connect registration block in the commit it was originally added
* Added new commit with rest of the grammar fixes.
When we are ready to merge, I plan to squash this down to a single commit. Let me know if that is a bad idea.

docs/Migration_Guide_Zephyr.md
sam-golioth: It should be in the migration guide. For those who were provisioning credentials in a custom way, especially those using offloaded DTLS sockets on the nrf9160, this is the information they need to move to the CP SDK.

docs/Migration_Guide_Zephyr.md
```diff
@@ -0,0 +1,595 @@
...
+    ```
+
+- Pass a pointer to the configuration when creating the Golioth client:
+
+    ```c
+    golioth_client_t client = golioth_client_create(&client_config);
+    ```
+
+- The callback that runs when Golioth connects is now registered using a
+  function call, and now passes a `golioth_client_event` into the
```
sam-golioth, line 245:
```suggestion
  function call, and the SDK now passes a `golioth_client_event` into the
```

sam-golioth:
> Thanks for the feedback, I made the following changes:
> 
> * Fix the table of contents in the two commits they were originally added
> * Fix the on_connect registration block in the commit it was originally added
> * Added new commit with rest of the grammar fixes.
>   When we are ready to merge, I plan to squash this down to a single commit. Let me know if that is a bad idea.

I'm good with the squash, I think that makes sense 👍

docs/Migration_Guide_Zephyr.md
sam-golioth: I think adding another example of the struct when using tags, like we have for psk and pki is enough. We don't need the full explanation in this guide.

szczys: Okay, I've added an example config.

sam-golioth (approved):

PR #306, opened 2024-01-25
zephyr: add mimxrt1060_evkb to sample tests

szczys:
Test alongside mimxrt1024_evk to see if the serial output is more stable during CI testing.


szczys:
Early testing shows this is not performing any better than the 1024 boards so we'll stick with those.

PR #307, opened 2024-01-26
client: support authentication by TLS credential tag

mniestroj:
Zephyr TLS subsystem uses tags for identifying credentials. Support that
method as alternative to passing credentials in plaintext. This gives
numeours benefits, like:
 * possibility to use credentials already provisioned and stored in secure
   storage
 * support for offloaded TLS sockets, like for nRF91 platform
 * support for external (application specific) method for provisioning
   credentials, without involvement of the Golioth SDK

TODO:
- [x] remove last commit 73e2dafc46a2 ("examples: zephyr: lightdb/get: configure PSK credentials in
      main()") which is for testing only


examples/zephyr/common/Kconfig
```diff
@@ -13,6 +13,22 @@ if GOLIOTH_SAMPLE_COMMON
 
 rsource "Kconfig.defconfig"
 
+choice GOLIOTH_SAMPLE_AUTH_TYPE
```
szczys, line 16: Should we make a default for this in https://github.com/golioth/golioth-firmware-sdk/blob/main/examples/zephyr/common/Kconfig.defconfig ?

include/golioth/client.h
```diff
@@ -63,6 +63,8 @@ enum golioth_auth_type {
     GOLIOTH_TLS_AUTH_TYPE_PSK,
     /// Authenticate with PKI certificates (CA cert, public client cert, private client key)
     GOLIOTH_TLS_AUTH_TYPE_PKI,
+    /// Authenticate with TLS credential tag (Zephyr specific)
```
szczys, line 66: Nit in commit message: `%s/numeours/numerous/g`

szczys (approved):
LGTM. One question on default symbols but great to have this addition, nice job!

sam-golioth (approved):

examples/zephyr/common/Kconfig
mniestroj: First options specified are default. Since PKI and PSK are exclusive based on `depends on GOLIOTH_AUTH_METHOD_PSK` and `depends on GOLIOTH_AUTH_METHOD_CERT` the only option that is non-default is newly introduced `GOLIOTH_SAMPLE_AUTH_TYPE_TAG`. So I don't thing we need to change anything.

include/golioth/client.h
mniestroj: Fixed!

PR #308, opened 2024-01-29
doxygen: generate sitemap

szczys:
generate a sitemap when Doxygen is built.

This feature is new in Doxygen 1.9.7 so the doxygen version used by CI needs to be downloaded and installed instead of using the package manager.


sam-golioth (approved):

beriberikix:
Yay! Looks like a sitemap to me 🗺️

PR #309, opened 2024-01-29
doxygen: fix root url used by sitemap

szczys:
The root url was specified by / but needs to be the actual url of the host site.


sam-golioth (approved):

beriberikix:
Looks like it should be fixed :+1:

PR #311, opened 2024-01-30
examples: zephyr: lightdb/get: configure PSK credentials in main()

mniestroj:
This is just for testing authentication by tag.


sam-golioth:
@mniestroj Are we good to close this in favor of #434 ?

mniestroj:
> @mniestroj Are we good to close this in favor of #434 ?

Yes, that is fine with me!

PR #314, opened 2024-01-31
examples: zephyr: lightdb: extend timeout

sam-golioth:
The lightdb examples were using a 1 second timeout for synchronous operations, which we can hit quite often on a cell network. Extending the timeout to 5 seconds should be more than sufficient.


szczys (approved):
Wow, good catch!

mniestroj (approved):

PR #323, opened 2024-02-01
Fix logging

sam-golioth:
If we got too many logs at once, we would hang both the coap_client thread and the logging thread. This fixes both, details in the commit messages.

### DIFF
--- a/src/coap_client.c
+++ b/src/coap_client.c
@@ -237,7 +237,11 @@ enum golioth_status golioth_coap_client_set(struct golioth_client *client,
     bool sent = golioth_mbox_try_send(client->request_queue, &request_msg);
     if (!sent)
     {
-        GLTH_LOGW(TAG, "Failed to enqueue request, queue full");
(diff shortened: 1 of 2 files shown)


src/coap_client.c
```diff
@@ -237,7 +237,11 @@ enum golioth_status golioth_coap_client_set(struct golioth_client *client,
     bool sent = golioth_mbox_try_send(client->request_queue, &request_msg);
     if (!sent)
     {
-        GLTH_LOGW(TAG, "Failed to enqueue request, queue full");
+        /* NOTE: Logging a message here when cloud logging is enabled can cause
+         *       a loop where the logging thread attempts to enqueue a message,
+         *       the mbox is full, so coap_client writes a log, which the
+         *       logging thread attempts to send to the cloud, and so on.
+         */
```
szczys, line 244: Excellent idea to leave a warning. I could see future me deciding this needed a log message.

szczys (approved):
Great catch, this is a tricky one.

PR #326, opened 2024-02-02
Fix decoding OTA manifest components

MarkoPura:
Add decoding of all manifest components (package, version, hash, size, URI, bootloader) for OTA updates. Necessary in order to correctly decode Releases with multiple Artifacts as a means to enable multi-MCU OTA updates.


include/golioth/ota.h
```diff
@@ -68,6 +68,13 @@ struct golioth_ota_component
     int32_t size;
     /// True, if the component is compressed and requires decompression
     bool is_compressed;
+    /// Artifact Hash
+    char hash[CONFIG_GOLIOTH_OTA_MAX_VERSION_LEN + 1];
```
sam-golioth, line 72: Should these constants be different? I'm also wondering if these should be dynamically allocated.

MarkoPura: They should, the hash is limited to 64 characters, I propose to leave it as a constant. Bootloader is either "mcuboot" or "default", so 7 characters. We can also discuss removing it from the payload (@hasheddan).
It makes the most sense to dynamically allocate the memory for the URI.

include/golioth/config.h
```diff
@@ -61,6 +61,18 @@
 #define CONFIG_GOLIOTH_OTA_MAX_VERSION_LEN 64
 #endif
 
+#ifndef CONFIG_GOLIOTH_OTA_MAX_COMPONENT_HASH_LEN
+#define CONFIG_GOLIOTH_OTA_MAX_COMPONENT_HASH_LEN 64
```
sam-golioth, line 65: Let's make the hash and bootloader name lengths constants in ota.h, rather than configurable values.

include/golioth/config.h
```diff
@@ -61,6 +61,18 @@
...
+#endif
+
+#ifndef CONFIG_GOLIOTH_OTA_MAX_COMPONENT_URI_LEN
+#define CONFIG_GOLIOTH_OTA_MAX_COMPONENT_URI_LEN 64
```
sam-golioth, line 69: We should keep this configurable, and also add it to Kconfig in `src/Kconfig`

include/golioth/ota.h
```diff
@@ -68,6 +68,13 @@ struct golioth_ota_component
     int32_t size;
     /// True, if the component is compressed and requires decompression
     bool is_compressed;
+    /// Artifact Hash
+    char hash[CONFIG_GOLIOTH_OTA_MAX_COMPONENT_HASH_LEN + 1];
+    /// Artifact uri (e.g. "/.u/c/main@1.2.3")
+    char uri [CONFIG_GOLIOTH_OTA_MAX_COMPONENT_URI_LEN + 1];
+    /// Artifact bootloader (e.g. "mcuboot")
+    char bootloader[CONFIG_GOLIOTH_OTA_MAX_COMPONENT_BOOTLOADER_NAME_LEN + 1];
```
sam-golioth, line 76: Adding these fields will increase RAM requirements non-trivially. As part of this PR, let's also change the default value for `CONFIG_GOLIOTH_OTA_MAX_NUM_COMPONENTS` to 1 (currently set at 4, which is excessive imo). Users can increase this config option if they need to do multi-component updates.

include/golioth/ota.h
```diff
@@ -18,6 +18,14 @@
...
 /// Maximum size of an OTA block, in bytes
 #define GOLIOTH_OTA_BLOCKSIZE 1024
+/// Size of a SHA256 of Artifact Binary in bytes
+#define GOLIOTH_OTA_COMPONENT_HASH_LEN 64
+/// Maximum size of Binary Detected Type - "mcuboot" or "default"
+#define GOLIOTH_OTA_MAX_COMPONENT_BOOTLOADER_NAME_LEN 7
+/// Maximum size of Relative URI to download binary
+#define GOLIOTH_OTA_MAX_COMPONENT_URI_LEN (CONFIG_GOLIOTH_OTA_MAX_PACKAGE_NAME_LEN + \
+                                           CONFIG_GOLIOTH_OTA_MAX_VERSION_LEN      + \
+                                           7)
```
sam-golioth, line 28: Let's add a comment for where we get 7 from

sam-golioth (approved):
Thanks for this Marko!

PR #327, opened 2024-02-02
examples: zephyr: nrf52840dk: increase packet pool

szczys:
The fw_update sample for the nrf52840 is still able to exhaust the packet pool at times. Bump the limit to accommodate.


examples/zephyr/fw_update/pytest/test_sample.py
```diff
@@ -10,11 +10,11 @@
 async def test_fw_update(shell, project, device, credentials_file, fw_info, release):
     # Read credentials
```
sam-golioth, line 11: Comment is out of order with the code now

examples/zephyr/fw_update/boards/mimxrt1024_evk.conf
```diff
@@ -2,3 +2,4 @@ CONFIG_GOLIOTH_SAMPLE_DHCP_BIND=y
 CONFIG_NET_L2_ETHERNET=y
 CONFIG_NET_DHCPV4=y
 CONFIG_BUILD_OUTPUT_HEX=y
+CONFIG_BOOT_DELAY=1
```
sam-golioth, line 5: This is specified in milliseconds, should it be 1000?

I also wonder if this should be part of the twister configuration instead of the board configuration, as we only need it for testing. On the other hand, this could be a material change in behavior between sample and test that might affect the accuracy of test results.

sam-golioth (approved):

PR #340, opened 2024-02-09
zephyr: add 'native_sim' with offloaded sockets

mniestroj:
Use:

``` sh
west update
west patch
```

to apply all downstream Zephyr patches.

Then build Zephyr example with `native_sim` or `native_sim_64` as a board.


scripts/west_commands/patch.py
```diff
@@ -0,0 +1,59 @@
...
+from west.commands import WestCommand, CommandError
+from west.manifest import MANIFEST_PROJECT_INDEX
+from west.app.project import _ProjectCommand
+
+class Patch(_ProjectCommand):
+    def __init__(self):
+        super().__init__(
+            'patch',
+            'patch west subprojects',
+            'Runs "git am patches/<project>/" on each of the specified projects.')
```
sam-golioth, line 16: Can we add some additional help information? Like, where should patches be located? How should the patch files be named? What additional metadata should be present in the manifest?

sam-golioth (approved):
Awesome to see this land!

Approved with the understanding that the actual patch files included in this PR will be updated to match the final changes merged to the Zephyr `main` branch before this PR is merged.

scripts/west_commands/patch.py
mniestroj: How about:
https://github.com/golioth/golioth-firmware-sdk/blob/6e9d68d4ee91225ac1977ec9f58c27b1304f255e/scripts/west_commands/patch.py#L9-L24

PR #341, opened 2024-02-09
Get wifi from gh secrets

sam-golioth:
This pulls WiFi credentials for each runner from GitHub secrets, instead of using prepositioned config files on the runners. With this is in place, we're almost ready to retire the configuration files, only the legacy `Test on Hardware` workflow still uses them.

The last commit is rather large, but it's making the same small change for each Zephyr sample.

Resolves golioth/firmware-issue-tracker#359


.github/workflows/hil_sample_zephyr.yml
```diff
@@ -129,6 +129,8 @@ jobs:
           hil_board: ${{ inputs.hil_board }}
           west_board: ${{ inputs.west_board }}
           GOLIOTH_API_KEY: ${{ secrets.PROD_CI_PROJECT_API_KEY }}
+          WIFI_SSID: ${{ secrets[format('{0}_WIFI_SSID', runner.name)] }}
```
szczys, line 132: nit: extra 'from' in the commit message:
> ci: make wifi creds from available to twister workflow

szczys (approved):

PR #352, opened 2024-02-21
examples: zephyr: cert_prov: reset at test start

sam-golioth:
The device can bootup and begin running the sample before the test starts. This can cause us to miss the startup message. Reset the device at the beginning of the test to ensure we start from a known point.

Fixes golioth/firmware-issue-tracker#449

### DIFF
--- a/examples/zephyr/certificate_provisioning/pytest/test_sample.py
+++ b/examples/zephyr/certificate_provisioning/pytest/test_sample.py
@@ -23,7 +23,8 @@ def subprocess_logger(result, log_msg):
 async def test_credentials(shell, project, device_name, device_port, certificate_cred):
 
     # Ensure there are no credentials currently stored on device
-
+    shell._device.clear_buffer()
+    shell._device.write('kernel reboot cold\n\n'.encode())
     shell._device.readlines_until(regex=".*Start certificate provisioning sample", timeout=90.0)
     sleep(5) # Time for fs bringup and in case nRF9160 reset loop reboot happens
     shell.exec_command(f'fs rm {FS_CRT_PATH}')


examples/zephyr/certificate_provisioning/pytest/test_sample.py
```diff
@@ -23,7 +23,8 @@ def subprocess_logger(result, log_msg):
 async def test_credentials(shell, project, device_name, device_port, certificate_cred):
 
     # Ensure there are no credentials currently stored on device
-
+    shell._device.clear_buffer()
+    shell._device.write('kernel reboot cold\n\n'.encode())
```
mniestroj, line 27: That is a workaround. This also means that such "reboot" invocation won't be possible with emulated targets like `native_sim`, which don't have "real" reboot mechanism. This is not a problem to this specific test, since this approach (reboot at the beginning of test) is already used in other samples, which is the main reason why `native_sim` cannot be used right away as testing platform.

Instead, in the long term we should figure a way to "start" running the firmware after already being attached to serial console. This can also use reboot as part of that implementation, but it should not be that explicit like invocation of Zephyr shell command. We could either attach to serial BEFORE flashing or even better (but probably more problematic to implement across different targets) to 1) flash but keep device/cpu stopped, 2) attach serial, 3) start/reset CPU to start execution.

mniestroj (approved):
Approving, since it solves HW issues already and we need to figure a better way anyway for all other samples, which can happen as the next step.

examples/zephyr/certificate_provisioning/pytest/test_sample.py
sam-golioth: Do you know if there's a way to do this in Twister? Right now part of the problem is that Twister flashes the board before calling pytest, so I'm not sure if there's a way for pytest to control execution other than through the serial port. We could add a `start` command or something to the serial console, but that would negatively impact the experience for users of the SDK.

mniestroj: I don't think there is such mechanism already in twister, at least I didn't encounter that yet.

Another workaround to consider would be starting the application (`main()` function) with a delay. There is a Kconfig option for that.

PR #366, opened 2024-03-04
Use `@retval` instead of `@return` when relevant

trond-snekvik:
`@retval` expects the first token to be a literal value, and renders it as code in the docs. This looks a little nicer than `@return` both in our generated docs and most code completion tools.

This PR changes the return value description to `@retval` for all functions that only list return values, but keeps the ones that still have a mix as-is.

[Example from Zephyr docs](https://docs.zephyrproject.org/latest/doxygen/html/group__memory__attr__heap.html#ga85301dfc19493a84e89f965ed793d576).

### Examples from Microsoft's C/C++ extension for VS Code:
`@return`:
Screenshot from 2024-03-04 12-32-01

`@retval`:
Screenshot from 2024-03-04 12-31-37

### DIFF
--- a/include/golioth/client.h
+++ b/include/golioth/client.h
@@ -204,8 +204,8 @@ bool golioth_client_wait_for_connect(struct golioth_client *client, int timeout_
 ///
 /// @param client The client handle
 ///
-/// @return GOLIOTH_OK Client started
-/// @return GOLIOTH_ERR_NULL Client handle invalid
+/// @retval GOLIOTH_OK Client started
+/// @retval GOLIOTH_ERR_NULL Client handle invalid
 enum golioth_status golioth_client_start(struct golioth_client *client);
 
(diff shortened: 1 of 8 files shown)


include/golioth/client.h
```diff
@@ -231,23 +231,23 @@ enum golioth_status golioth_client_stop(struct golioth_client *client);
 /// @param client The handle of the client to destroy
 void golioth_client_destroy(struct golioth_client *client);
 
-/// Returns whether the client is currently running
+/// retvals whether the client is currently running
```
hasheddan, line 234: find + replace going rogue 😉 
```suggestion
/// Returns whether the client is currently running
```

trond-snekvik: :see_no_evil:

sam-golioth (approved):

PR #367, opened 2024-03-04
ci: erase nRF9160 at end of tests

szczys:
Enable nRF9160dk CI testing with an erase step as the final (required) step of the workflow. This ensures that after a test a dev board will not continue to be connected to the cellular network, accruing data usage charges.

### DIFF
--- a/.github/workflows/hil_sample_zephyr.yml
+++ b/.github/workflows/hil_sample_zephyr.yml
@@ -180,8 +180,27 @@ jobs:
           hil_board: ${{ inputs.hil_board }}
         run: |
           source /opt/credentials/runner_env.sh
+          export PATH=$PATH:/opt/SEGGER/JLink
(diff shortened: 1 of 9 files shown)


sam-golioth:
Would it be straightforward to also erase the other boards?

sam-golioth:
Let's investigate why there are so many failures on the 9160 twister tests.

sam-golioth:
> Would it be straightforward to also erase the other boards?

@szczys latest changes look good, but still waiting on this

szczys:
> Let's investigate why there are so many failures on the 9160 twister tests.

szczys:
> > Would it be straightforward to also erase the other boards?
> 
> @szczys latest changes look good, but still waiting on this

I've added erase commands for the rest of the boards and CI has passed.

szczys:
Oops, clicked close by mistake

sam-golioth (approved):

PR #373, opened 2024-03-06
CI: HIL: test OTA service

szczys:
Add CI test for Over-the-Air (OTA) firmware update service.


szczys:
There are some test stability issues that need addressing (like nrf52840 failling connection test) but I don't think they are directly related to changes in this PR so I'm requesting review now.

szczys:
@sam-golioth as discussed I added a fixture to roll-back the releases after each test. I ran the tests individually, and all together to verify it behaves as expected.

tests/hil/tests/ota/test_ota.py
```diff
@@ -0,0 +1,189 @@
...
+GOLIOTH_OTA_STATE_CNT = len(golioth_ota_state)
+GOLIOTH_OTA_REASON_CNT = len(golioth_ota_reason)
+TEST_BLOCK_CNT = 42
+
+pytestmark = pytest.mark.anyio
+
+@pytest.fixture(autouse=True, scope="module")
+async def setup(board, device):
+    # Delay to give NVS time to initialize
+    time.sleep(4)
```
sam-golioth, line 41: This is handled by `board.py` now.

tests/hil/tests/ota/test_ota.py
```diff
@@ -0,0 +1,189 @@
...
+            if a.version == DUMMY_VER_OLDER and a.package == UPDATE_PACKAGE:
+                artifacts["test_manifest"] = a
+            elif a.version == DUMMY_VER_SAME and a.package == UPDATE_PACKAGE:
+                artifacts["test_blocks"] = a
+            elif a.version == DUMMY_VER_UPDATE and a.package == UPDATE_PACKAGE:
+                artifacts["test_reasons"] = a
+
+    assert len(artifacts) == 3
+
+    yield artifacts
```
sam-golioth, line 87: Can this just be `return` since we're not doing any cleanup?

tests/hil/tests/ota/test_ota.py
```diff
@@ -0,0 +1,189 @@
...
+
```
sam-golioth, line 88: NIT: extra whitespace

tests/hil/tests/ota/test_ota.py
```diff
@@ -0,0 +1,189 @@
...
+    for r in releases:
+        try:
+            await project.releases.delete(releases[r].id)
+        except:
+            print(f"Failed to delete release: {releases[r].id}")
+            pass
+
+@pytest.fixture(scope="function")
+async def releases_teardown(project, releases):
```
sam-golioth, line 109: This is running before each test, instead of after. The effect is probably the same, but the name is a little misleading.

Also, since this isn't actually returning anything and all the tests are using this fixture, we can set `autouse=True` in the decorator, and remove the `releases_teardown` argument from each test.

sam-golioth (commented):
Can we also add a test for decoding multiple components in a manifest?

szczys (commented):
@sam-golioth I've addressed all feedback, and added a test for multiple artifacts.

tests/hil/tests/ota/test_ota.py
```diff
@@ -0,0 +1,198 @@
...
+    assert None != board.wait_for_regex_in_line(f'main.bootloader: {b_info["type"]}', timeout_s=2)
+
+
+async def test_multiple_artifacts(artifacts, board, project, releases):
+    await project.releases.rollout_set(releases["test_multiple"].id, True)
+
+    assert None != board.wait_for_regex_in_line('Manifest received', timeout_s=30)
+    assert None != board.wait_for_regex_in_line('Manifest successfully decoded', timeout_s=2)
+    assert None != board.wait_for_regex_in_line('Found main component', timeout_s=30)
+    assert None != board.wait_for_regex_in_line('Found walrus component', timeout_s=2)
```
sam-golioth, line 153: Any reason to not also check the manifest contents?

szczys: I've updated to include this. All tests passed, this is ready for review.

sam-golioth (approved):

PR #374, opened 2024-03-07
CI: modustoolbox: fix download from Google Drive

szczys:
Update `gdown` python package to latest to fix broken Google Drive download.

### DIFF
--- a/.github/workflows/lint_build_unit_test.yml
+++ b/.github/workflows/lint_build_unit_test.yml
@@ -151,7 +151,7 @@ jobs:
     - name: Download and install ModusToolbox 2.4
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/lint_build_unit_test.yml
```diff
@@ -151,7 +151,7 @@ jobs:
     - name: Download and install ModusToolbox 2.4
       shell: bash
       run: |
-        pip install click gdown==4.6.0 cryptography==41.0.7 intelhex cbor
+        pip install click gdown cryptography==41.0.7 intelhex cbor
```
sam-golioth, line 154: Let's pin this to `5.1.0`. We had to pin this originally (in 011d6ee9018ebeb4371d07a79c5a0361263fb0fc) because an update broke the download

sam-golioth (approved):

PR #385, opened 2024-03-11
Update runtime credentials without reboot

szczys:
Allow runtime credentials to be used immediately after being updated, without the need for a reboot.

Caveats:
- The ESP-IDF implementation of this requires that credentials not be present in NVS, otherwise the logic for updating them without reboot may be missed at runtime. I discussed this with @sam-golioth and we felt this meets the needs of CI which is the primary driver of this change
- In Zephyr, an error will be thrown whenever credentials are updated before the golioth client is client is created. This is because the SDK tries to set the tls_credential and finds it is already present, causing the -17 error as shown below.

```
uart:~$ [00:00:04.777,557] <err> golioth_coap_client_zephyr: Failed to register PSK ID: -17
```

Resolves https://github.com/golioth/firmware-issue-tracker/issues/454

### DIFF
--- a/examples/zephyr/common/settings_golioth.c
+++ b/examples/zephyr/common/settings_golioth.c
@@ -10,6 +10,7 @@ LOG_MODULE_REGISTER(golioth_sample_settings, CONFIG_GOLIOTH_DEBUG_DEFAULT_LOG_LE
 #include <errno.h>
 #include <golioth/client.h>
 #include <zephyr/init.h>
+#include <zephyr/net/tls_credentials.h>
 #include <zephyr/settings/settings.h>
 
 /*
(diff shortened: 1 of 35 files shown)


mniestroj (approved):

sam-golioth:
> In Zephyr, an error will be thrown whenever credentials are updated before the golioth client is client is created. This is because the SDK tries to set the tls_credential and finds it is already present, causing the -17 error as shown below.

Can we attempt to delete the credential before setting it in the SDK?

sam-golioth:
We'll need to rebase this and update tests that have been merged in since this first went up. We'll also need to make changes to the Linux Board class in pytest-hil. Right now we start running the application when board.reset() is called. We can just move the contents of the function Board.set_psk_credentials().

examples/zephyr/common/settings_golioth.c
```diff
@@ -10,8 +10,10 @@ LOG_MODULE_REGISTER(golioth_sample_settings, CONFIG_GOLIOTH_DEBUG_DEFAULT_LOG_LE
 #include <errno.h>
 #include <golioth/client.h>
 #include <zephyr/init.h>
+#include <zephyr/net/tls_credentials.h>
 #include <zephyr/settings/settings.h>
 
+
```
mniestroj, line 16: Superfluous newline.

mniestroj (approved):

sam-golioth (approved):

szczys:
@sam-golioth I've made the changes as discussed. Please review this commit to ensure it aligns with your suggestion of deleteing the TLS cred before setting it in the SDK https://github.com/golioth/golioth-firmware-sdk/pull/385/commits/65b25838a1a75f1e8795477aff31e7da17d2183c

PR #389, opened 2024-03-12
workflows: esp-idf: erase boards after tests

szczys:
If a board had stored credentials when a test runs, the new credentials may not be used immediatel in some cases. For instance, if a WiFi-based board has already established a connection when new credentials are set, they will not be immediately used since a connetion is currently in place.

This commit erases board flash at the end of test runs to ensure all applications will be provisioned with the expected AP/Golioth device credentials the next time a test runs.

### DIFF
--- a/.github/workflows/hil_sample_esp-idf.yml
+++ b/.github/workflows/hil_sample_esp-idf.yml
@@ -94,3 +94,12 @@ jobs:
               --wifi-psk ${{ secrets[format('{0}_WIFI_PSK', runner.name)] }}    \
               --timeout=600
           done
+      - name: Erase flash
+        if: always()
+        shell: bash
+        env:
+          hil_board: ${{ inputs.hil_board }}
+        run: |
+          source /opt/credentials/runner_env.sh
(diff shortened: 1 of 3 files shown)


szczys:
@sam-golioth asked me to cycle test this change to ensure that erasing the flash didn't alter the timing of the test the next time they run. Do do so, I created a branch that only tested the ESP boards and I added a second set of steps after the erase step to USB power cycle and re-run tests. This way the same board is tested, erased, and tested again all in one action.

These all passed except [one test](https://github.com/golioth/golioth-firmware-sdk/actions/runs/8271441077/job/22631585979) that was already fixed but [that commit](https://github.com/golioth/golioth-firmware-sdk/commit/e1ce5bd1917efbd01e00867d80b1117dbcad753b) didn't make it into my test branch. Here are the workflows:

- https://github.com/golioth/golioth-firmware-sdk/actions/runs/8271441041
- https://github.com/golioth/golioth-firmware-sdk/actions/runs/8271441077

sam-golioth (approved):

PR #391, opened 2024-03-12
 zephyr: backend logging: enable/disable using a context flag

szczys:
The Zephyr port was using log_backend_enable()/disable() to prevent logs from generating CoAP events when not connected to Golioth. This interferes with applications using the same functions to enable or disable backend logging at runtime.

- test for null context before checking context members
- add an `enabled` flag to the backend context which is used to determine if a log should be processed
- format the code (I'm unsure how this file was missed during the giant code reformat PR awhile ago

### DIFF
--- a/examples/zephyr/certificate_provisioning/pytest/test_sample.py
+++ b/examples/zephyr/certificate_provisioning/pytest/test_sample.py
@@ -25,14 +25,14 @@ async def test_credentials(shell, project, device_name, device_port, certificate
     # Ensure there are no credentials currently stored on device
     shell._device.clear_buffer()
     shell._device.write('kernel reboot cold\n\n'.encode())
-    shell._device.readlines_until(regex=".*Start certificate provisioning sample", timeout=90.0)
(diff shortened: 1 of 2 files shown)


sam-golioth (approved):
Awesome, thanks for tracking this down @szczys! 

I think I (erroneously) skipped the whole `port` folder when reformatting. Glad to see it happening now 👍

sam-golioth:
The certificate provisioning sample on MIMXRT1024 failed 5 times in a row. Let's investigate that failure before merging this.

szczys:
This approach caused another issue on the mimxrt1024_evk. Now that the Golioth logging backend is autostarting, for this board it is the first backend to be initialized (before the shell backend can be initialized). Since the Golioth logging backend returns without logging when there is no connection in place, this meant that there were a few seconds before the shell backend was initialized so the logs during that early start-up were not printed.

Closing this PR and starting a new branch to track this work: https://github.com/golioth/golioth-firmware-sdk/tree/szczys/zephyr-logging-backend-rethink-flags

PR #395, opened 2024-03-13
ESP-IDF Stream Example

rashmivaidya72:
NOTE:  The Zephyr LightDB Stream example uses the `struct sensor_value` from the Zephyr SDK. As this structure is not available outside of Zephyr, I have included its definition in the ESP-IDF example. Also, the Zephyr Example allows for obtaining temperature values directly from a temperature sensor if the platform supports it. However, the ESP-IDF example always generates fake temperature values.


examples/esp_idf/lightdb_stream/README.md
```diff
@@ -0,0 +1,209 @@
...
+I (50119) lightdb_stream: Sending temperature 24.500000
+I (50429) lightdb_stream: Temperature successfully pushed
+I (55119) lightdb_stream: Sending temperature 25.000000
+I (55239) lightdb_stream: Temperature successfully pushed
+```
+
+### Monitor temperature value over time
+
+Device sends temperature measurements every 5s and updates `/sensor/temp`
+resource in LightDB Stream. Current value can be fetched using following
```
sam-golioth, line 112:
```suggestion
resource in LightDB Stream. The current value can be fetched using the following
```

examples/esp_idf/lightdb_stream/README.md
```diff
@@ -0,0 +1,209 @@
...
+You may have to set target based on the ESP32 chip you are using.
+For example, if you are using ESP32-C3, enter this:
+
+```
+idf.py set-target esp32c3
+```
+
+Next, `cd` to the lightdb_stream example where you can build/flash/monitor:
+
+Make sure that CONFIG_GOLIOTH_LIGHTDB_STREAM is set by going into the interactive menuconfig
```
sam-golioth, line 35: This is set in `sdkconfig.defaults` so we don't need instructions on setting it through menuconfig here.

examples/esp_idf/lightdb_stream/main/app_main.c
```diff
@@ -0,0 +1,216 @@
...
+#include <string.h>
+
+#define TAG "lightdb_stream"
+
+struct golioth_client *client;
+static SemaphoreHandle_t _connected_sem = NULL;
+
+
+/**
+ * Definition taken from Zephyr SDK
```
sam-golioth, line 22: We don't need to match anything Zephyr specific here. The goal behind having examples for each platform (instead of shared examples) is to show how each service can be integrated into that platform, including any of that platform's specific features or weirdness. For this example, we can either just `printf` a float value directly, or even just use integer values (e.g. 20, 21, 22, etc.).

examples/esp_idf/lightdb_stream/main/app_main.c
```diff
@@ -0,0 +1,192 @@
...
+{
+    static int counter = 0;
+
+    /* generate a temperature from 20 deg to 30 deg, with 0.5 deg step */
+
+    *temp = 20.0 + (float)counter / 2;
+
+    counter = (counter + 1) % 20;
+
+    return 0;
```
sam-golioth, line 42: Since this always returns `0`, we can just have this return `temp` instead of an error code and we don't have to pass in the reference.

examples/esp_idf/lightdb_stream/main/app_main.c
```diff
@@ -0,0 +1,192 @@
...
+        err = get_temperature(&temp);
+        if (err)
+        {
+            vTaskDelay(1000 / portTICK_PERIOD_MS);
+            continue;
+        }
+
+        GLTH_LOGI(TAG, "Sending temperature %f", temp);
+
+        temperature_push_sync(&temp);
```
sam-golioth, line 175: Is there any reason to pass this by reference instead of by value?

rashmivaidya72: No reason. This was how the zephyr example did it and I followed the same method. I'll change it.

examples/esp_idf/lightdb_stream/main/app_main.c
rashmivaidya72: Right

sam-golioth:
@rashmivaidya72 The lightdb stream sample test failed on the esp32s3: https://github.com/golioth/golioth-firmware-sdk/actions/runs/8424019673/job/23067520531?pr=395

rashmivaidya72:
> @rashmivaidya72 The lightdb stream sample test failed on the esp32s3: https://github.com/golioth/golioth-firmware-sdk/actions/runs/8424019673/job/23067520531?pr=395

@sam-golioth This seems to be a timing issue. The test expects 20.0, but the DUT has already published 20.0 and 20.5, causing the assertion to fail.
FAILED examples/esp_idf/lightdb_stream/pytest/test_sample.py::test_lightdb_stream - assert 20.5 == 20.0

If I remove the check for Golioth client connected, the test speeds up and the assertion succeeds.

examples/esp_idf/lightdb_stream/pytest/test_sample.py
```diff
@@ -0,0 +1,36 @@
...
+
+async def test_lightdb_stream(board, device):
+    time.sleep(2)
+
+    # Set Golioth credential
+    golioth_cred = (await device.credentials.list())[0]
+    board.set_golioth_psk_credentials(golioth_cred.identity, golioth_cred.key)
+
+    # Reset board
+    board.reset()
```
sam-golioth, line 25: We can remove this now that https://github.com/golioth/golioth-firmware-sdk/pull/385 is merged.

sam-golioth:
We need to rebase this PR and fix the failing CI tests.

sam-golioth:
Seeing intermittent failures on the CI test: https://github.com/golioth/golioth-firmware-sdk/actions/runs/8546397122/job/23416975750?pr=395#step:7:244

sam-golioth:
@mniestroj @szczys This one is ready for a review

examples/esp_idf/stream/pytest/test_sample.py
```diff
@@ -0,0 +1,36 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import contextlib
+import logging
+import pytest
+import time
+import yaml
+import datetime
+import re
```
mniestroj, line 11: 3 unused imports.

examples/esp_idf/stream/pytest/test_sample.py
```diff
@@ -0,0 +1,36 @@
...
+LOGGER = logging.getLogger(__name__)
+
+pytestmark = pytest.mark.anyio
+
+async def test_stream(board, device):
+    time.sleep(2)
+
+    # Set Golioth credential
+    golioth_cred = (await device.credentials.list())[0]
+    board.set_golioth_psk_credentials(golioth_cred.identity, golioth_cred.key)
```
mniestroj, line 22: Not a blocker for this PR, but would be good to use the same method for `examples/zephyr/stream` as well (instead of manually calling `settings set golioth/...` commands, since we already have abstraction for that. Just to keep implementation differences as minimal as possible.

mniestroj: Another question: what about settings WiFi credentials? Is ESP-IDF port different in this regard compared to Zephyr? (In Zephyr's pytest script there is an explicit configuration of WiFi credentials)

examples/esp_idf/stream/pytest/test_sample.py
```diff
@@ -0,0 +1,36 @@
...
+    # Verify temp messages
+
+    lower_temp = 20.0
+    upper_temp = 21.5
+    async with contextlib.aclosing(device.stream.iter()) as stream_iter:
+        async for value in stream_iter:
+            LOGGER.info("ts: {0}, temp: {1}".format(value['timestamp'], value['data']['sensor']['temp']))
+            assert (value["data"]["sensor"]["temp"] >= lower_temp and value["data"]["sensor"]["temp"] <= upper_temp)
+            if value["data"]["sensor"]["temp"] == upper_temp:
+                break
```
mniestroj, line 36: In b85446ba73a4 ("examples: zephyr: stream: use CBOR over JSON") we removed `sensor` from the path where data goes. Should we remove it here as well to keep samples consistent for all platforms?

examples/esp_idf/stream/pytest/test_sample.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # Verify temp messages
+
+    lower_temp = 20.0
+    upper_temp = 21.5
+    async with contextlib.aclosing(device.stream.iter()) as stream_iter:
+        async for value in stream_iter:
+            LOGGER.info("ts: {0}, temp: {1}".format(value['timestamp'], value['data']['sensor']['temp']))
+            assert (value["data"]["sensor"]["temp"] >= lower_temp and value["data"]["sensor"]["temp"] <= upper_temp)
+            if value["data"]["sensor"]["temp"] == upper_temp:
```
szczys, line 35:
```suggestion
            LOGGER.info("ts: {0}, temp: {1}".format(value['timestamp'], value['data']['temp']))
            assert (value["data"]["temp"] >= lower_temp and value["data"]["temp"] <= upper_temp)
            if value["data"]["temp"] == upper_temp:
```

examples/esp_idf/stream/pytest/test_sample.py
sam-golioth: > would be good to use the same method for examples/zephyr/stream as well

These functions are part of our `pytest-hil` pytest plugin, which isn't compatible with the Twister pytest plugins, so we can't use them for the Zephyr tests.

> what about settings WiFi credentials? Is ESP-IDF port different in this regard compared to Zephyr?

WiFi credentials are set by the `pytest-hil` plugin's `board` fixture. The `board` fixture doesn't set PSK credentials because those are generate at run-time by the test using the `python-golioth-tools` pytest plugin's `device` fixture.

mniestroj: Thanks for explanation!

szczys (approved):

mniestroj (approved):

PR #397, opened 2024-03-14
Use flags to control Golioth logging backend

szczys:
The Zephyr port was using log_backend_enable()/disable() to prevent logs from generating CoAP events when not connected to Golioth. This interferes with applications using the log_backend_activate()/deactivate() functions to enable or disable backend logging at runtime.

This commit adds an `enabled` flag to the backend context which is used to determine if a log should be processed. This way, the SDK can disable/enable backend logging when the connection status changes, while retaining the ability for applications to control the backend through normal Zephyr logging management functions.

The first time a connection to Golioth is established, the backend will be automatically activated. If your application needs to deactivate this backend at runtime, do so only after the first connection is established.

### DIFF
--- a/port/zephyr/golioth_log_zephyr.c
+++ b/port/zephyr/golioth_log_zephyr.c
@@ -31,6 +31,7 @@ struct cbpprintf_ctx
 struct golioth_log_ctx
 {
     struct golioth_client *client;
+    bool enabled;
     bool panic_mode;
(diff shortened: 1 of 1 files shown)


szczys:
The nrf9160 failures in CI look like connectivity issues.

sam-golioth (approved):

PR #398, opened 2024-03-14
tests: hil: add rpc tests

sam-golioth:
Resolves golioth/firmware-issue-tracker#321


sam-golioth:
The ESP-IDF RPC sample tests are unstable in this PR. I don't immediately know why that might be, but I'm going to investigate (mostly commenting to remind myself).

sam-golioth:
> The ESP-IDF RPC sample tests are unstable in this PR. I don't immediately know why that might be, but I'm going to investigate (mostly commenting to remind myself).

This is resolved in #400 and is unrelated to the tests introduced here. This PR is ready for review.

tests/hil/tests/rpc/test.c
```diff
@@ -0,0 +1,150 @@
...
+static enum golioth_rpc_status on_no_response(zcbor_state_t *request_params_array,
+                                              zcbor_state_t *response_detail_map,
+                                              void *callback_arg)
+{
+    size_t arg_count = 0;
+    while (!zcbor_list_or_map_end(request_params_array))
+    {
+        arg_count++;
+        int err = zcbor_any_skip(request_params_array, NULL);
+        if (!err)
```
szczys, line 23: This is a bit hard to understand since [zcbor doesn't document the return value of the function](https://github.com/NordicSemiconductor/zcbor/blob/75fa5b2703f921433eee32475deed0e9790c968b/include/zcbor_decode.h#L266-L276). Does it return true if the skip was successful?

szczys (approved):

tests/hil/tests/rpc/test.c
sam-golioth: Yeah, this should be probably be `bool ok` instead of `int err`, as in other places where we call zcbor APIs. Will update.

PR #400, opened 2024-03-15
examples: esp-idf: rpc: wait for RPC registered before calling in test

sam-golioth:
Some failures in CI were caused by the test script attempting to call the RPC before the device had established an observation on the RPC endpoint. This commit adds a new message that the device prints when it has established the observation, and updates the test script to wait for that message.

Fixes golioth/firmware-issue-tracker#487


szczys (approved):

sam-golioth:
@szczys my previous approach failed CI. I've reworked this PR and it could use another review.

szczys (approved):

PR #402, opened 2024-03-18
Fix building without Newlib Libc with NCS

sense-Jo:
When using NCS, builds for targets which do not use Newlib Libc would fail since commit 0f33246.
In particular, building e.g. the `hello` sample for `qemu_x86` would fail.


sam-golioth:
Thanks @sense-Jo! We noticed this issue this morning as well. We'll be taking a different approach to solving this, but I appreciate the contribution!

PR #407, opened 2024-03-19
rpc: add GOLIOTH_RPC_MAX_RESPONSE_LEN

szczys:
Add Kconfig setting for the maximum RPC response length. This value will determine how many bytes are available to encode the response_detail_map passed to the RPC callback, used to return data to the server at the end of the RPC run.

This memory was previously allocated on the stack. To avoid stack overflow when adjusting this new symbol, the memory is now statically allocated to keep it out of the stack. A planned future improvement will make this dynamically allocated.


src/rpc.c
```diff
@@ -122,7 +122,7 @@ static void on_rpc(struct golioth_client *client,
     }
 
     /* Start encoding response */
-    uint8_t response_buf[256];
+    static uint8_t response_buf[CONFIG_GOLIOTH_RPC_MAX_RESPONSE_LEN];
```
mniestroj, line 125: Please add `static` in a separate commit, for increased visibility in git log.

sam-golioth (approved):

PR #409, opened 2024-03-19
CI: power USB hub on before test and off afterward

szczys:
- Update workflows to turn USB hub on at the start of a test and off at the end of it.
- Update docker images with USB hub control that uses parameters for on/off/cycle control.

resolves https://github.com/golioth/firmware-issue-tracker/issues/389

### DIFF
--- a/.github/workflows/hil_sample_esp-idf.yml
+++ b/.github/workflows/hil_sample_esp-idf.yml
@@ -58,7 +58,7 @@ jobs:
     runs-on: [ is_active, "has_${{ inputs.hil_board }}" ]
 
     container:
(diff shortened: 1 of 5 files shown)


szczys:
When observing the hubs while CI for this PR ran I noticed that the data connectivity lights go off, but some boards didn't power down. Further investigation shows that for the model of hubs we have, only 2 of the 7 ports are capable of switching the power lines.

I've ordered a different hub model to test and will submit this PR for review when I have confirmed that device as replacement for our runner setups.

sam-golioth (approved):

sam-golioth (approved):

PR #414, opened 2024-03-21
ci: hil: set CoAP URI in correct sdkconfig.defaults files

hasheddan:
Updates the HIL esp-idf samples workflow to set the CoAP URI in the sdkconfig.defaults file of each respective sample, rather than the HIL connection test.


hasheddan:
Replaced by https://github.com/golioth/golioth-firmware-sdk/pull/416

PR #415, opened 2024-03-21
Capture code coverage for Linux HIL integration tests

sam-golioth:
This generates code coverage information during the Linux integration "HIL" tests. The CI workflow then uses `lcov` and `genhtml` to generate coverage reports. A detailed HTML report is saved as an artifact of the GitHub Actions workflow. If the workflow was triggered from a pull request, then the workflow will also comment on the PR with the overall coverage statistic.

We are capturing code coverage information from the Linux port mostly because it's easier. We don't have to worry about the code instrumentation affecting performance or bloating the size of the binary to the point it won't fit in the device flash. We also don't have to come up with a way to pull the code coverage reports off of the DUT.

We are calculating code coverage against the `src` folder, as that is the core part of our SDK, and because the code coverage numbers generated there transfer well to other ports that our customers would actually run on their hardware (i.e. generating code coverage information against the `port` folder only tells us about coverage for the Linux port).

Note that these numbers are only relevant for the libcoap ports (ESP-IDF, ModusToolbox, Linux), because the Zephyr port uses a different CoAP stack. We will capture code coverage information for the Zephyr port in a similar way against the `native_sim` target later. We will also combine the code coverage reports with coverage information from the unit tests to get a fuller picture.

Resolves golioth/firmware-issue-tracker#35


szczys (approved):
This is great!

PR #421, opened 2024-03-25
ci: support workflow_dispatch on HIL sub-workflows

hasheddan:
Adds workflow_dispatch trigger to HIL sub-workflows such that they can be invoked individually.

### DIFF
--- a/.github/workflows/hil_sample_esp-idf.yml
+++ b/.github/workflows/hil_sample_esp-idf.yml
@@ -1,6 +1,23 @@
 name: ESP-IDF Hardware-in-the-Loop Sample Tests
 
 on:
+  workflow_dispatch:
+    inputs:
+      hil_board:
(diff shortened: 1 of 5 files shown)


.github/workflows/hil_sample_esp-idf.yml
```diff
@@ -1,6 +1,23 @@
...
+        type: string
+      api-url:
+        required: true
+        type: string
+      api-key-id:
+        required: true
+        type: string
+      coap_gateway_url:
+        required: true
+        type: string
```
hasheddan, line 20: mirrored the no defaults / no description in each of these for `workflow_call` -- happy to add them if helpful! Imagine that folks invoking these are going to be quite familiar with the parameters without descriptions / defaults tho

sam-golioth (approved):

PR #431, opened 2024-03-28
Add sample_credentials module

rashmivaidya72:


examples/esp_idf/common/sample_credentials.c
```diff
@@ -0,0 +1,78 @@
...
+{
+#ifdef CONFIG_GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS
+
+    _golioth_client_config.credentials.psk.psk_id = CONFIG_GOLIOTH_SAMPLE_PSK_ID;
+    _golioth_client_config.credentials.psk.psk_id_len = strlen(CONFIG_GOLIOTH_SAMPLE_PSK_ID);
+    _golioth_client_config.credentials.psk.psk = CONFIG_GOLIOTH_SAMPLE_PSK;
+    _golioth_client_config.credentials.psk.psk_len = strlen(CONFIG_GOLIOTH_SAMPLE_PSK);
+
+    // Initialize WiFi and wait for it to connect
+    printf("calling wifi_init with %s and %s\n", CONFIG_GOLIOTH_SAMPLE_WIFI_SSID, CONFIG_GOLIOTH_SAMPLE_WIFI_PSK);
```
sam-golioth, line 46: We shouldn't be doing anything with WiFi in this module. It probably makes sense to commonize the bits of the sample that get WiFi credentials, but that can be in a separate ticket/PR and a separate module. We also don't necessarily want to tie the runtime vs hardcoded choice for WiFi settings to the same choice for Golioth credentials. For example, the user could select hardcoded PSK but runtime WiFi credentials (maybe unlikely, but possible).

port/esp_idf/components/golioth_sdk/Kconfig
sam-golioth, line 1: We should add this Kconfig options into a Kconfig file in `examples/esp_idf/common` and source that file from each of the examples. That way these options, which are just for our examples, aren't available to the customer applications (they should be using some sort of secure storage for their credentials).

examples/esp_idf/hello/main/app_main.c
sam-golioth, line 1: We'll need to make similar changes to the other examples.

sam-golioth (commented):
Just a note that we'll need to fix the linter failures as well.

examples/esp_idf/common/Kconfig
```diff
@@ -0,0 +1,29 @@
...
+# Copyright (C) 2024 Golioth, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+menu "Common helper configurations shared across multiple samples"
+
+config GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS
+    bool "Hardcoded credentials"
+    default n
```
sam-golioth, line 11:
```suggestion
```
No need to specify `n`, it's already the default.

examples/esp_idf/common/Kconfig
```diff
@@ -0,0 +1,29 @@
...
+    help
+        Use hardcoded credentials in samples
+
+if GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS
+
+config GOLIOTH_SAMPLE_PSK_ID
+    string "PSK ID"
```
sam-golioth, line 18:
```suggestion
    string "Golioth PSK-ID"
```

examples/esp_idf/common/Kconfig
```diff
@@ -0,0 +1,29 @@
...
+    help
+        Defines PSK ID used during DTLS handshake with Golioth server.
+
+config GOLIOTH_SAMPLE_PSK
+    string "PSK"
```
sam-golioth, line 23:
```suggestion
    string "Golioth PSK"
```

sam-golioth (approved):

PR #435, opened 2024-04-03
zephyr: add rak5010 board support

szczys:
- [x] add board files to all samples
- [x] add build/flash documentation to all README files

## Sample testing:

- [x] hello
- [x] lightdb set
- [x] lightdb get
- [x] lightdb observe
- [x] lightdb delete
- [x] lightdb stream
- [x] logging
- [x] rpc
- [x] settings
- [x] ota
- [x] certs
- [x] golioth basics

resolves https://github.com/golioth/firmware-issue-tracker/issues/491


examples/zephyr/common/net_connect.c
```diff
@@ -48,6 +48,17 @@ void net_connect(void)
...
+    {
+        LOG_INF("Bringing up network interface");
+        int ret = net_if_up(iface);
+        if ((ret < 0) && (ret != -EALREADY))
+        {
+            LOG_ERR("Failed to bring up network interface: %d", ret);
+            return;
+        }
+    }
+
```
szczys, line 61: @mniestroj I would be particularly interested if you have any advice regarding this change. The network interface for the rak5010 doesn't start automatically, I've found I need to call `net_if_up()` where we previously did not. If there is a better approach, I'd be happy to make a change. Thanks!

examples/zephyr/certificate_provisioning/boards/rak5010_nrf52840.conf
```diff
@@ -0,0 +1,26 @@
+CONFIG_UART_ASYNC_API=y
+CONFIG_MODEM_CELLULAR_APN="internet"
+
+# USB
+CONFIG_USB_DEVICE_STACK=y
+CONFIG_USB_DEVICE_PRODUCT="Zephyr USB console sample"
```
sam-golioth, line 6: Should we change this name?

sam-golioth:
Can we add `rak5010_nrf52840` to all of the twister tests, including the ones that run on pytest. Then we should add another commit that has CI build the tests, but mark it as `run_tests: false` in the workflow yaml, so we can get build test coverage on all of the samples even though we don't have the twister tests working yet. Same as we do for `qemu_x86` and `esp32_devkitc_wrover`.

sam-golioth (approved):
🐳

PR #438, opened 2024-04-05
bump libcoap to v4.3.4a

mniestroj:
This is the latest tagged release.


sam-golioth (approved):
Are there any changes that are particularly relevant for us or just updating?

mniestroj:
> Are there any changes that are particularly relevant for us or just updating?

There is NULL pointer dereference bug fixed in this version. I was running into it on my Linux machine. This is the commit: https://github.com/obgm/libcoap/commit/ae2cc9ffef56245177b7682f44b66da648d05aea

PR #443, opened 2024-04-12
coap_client: propagate response errors to sychronous API calls

mniestroj:
So far there was no way to pass error codes (besides timeout) from
synchronous (CoAP) requests. Add `enum golioth_status` to the
`golioth_coap_request_msg_t` to make it possible to return non-success
status codes.

When response from server was received with error code, synchronous API
calls were still retuning `GOLIOTH_OK` status.

Propagate responses with error codes to the application as error status
`GOLIOTH_ERR_*`.


src/coap_client_zephyr.c
```diff
@@ -178,6 +204,8 @@ static int golioth_coap_cb(struct golioth_req_rsp *rsp)
     {
         if (req->request_complete_event)
         {
+            *req->status = golioth_err_to_status(rsp->err);
```
sam-golioth, line 207: Do we need to make a similar change for libcoap?

mniestroj: Yes, that should be the case. Though, I am not sure yet how the implementation would look like.

sam-golioth: Ok, let's make the change for both platforms at the same time.

mniestroj: This PR was updated to cover all platforms, so please take another look @sam-golioth.

sam-golioth (approved):

PR #444, opened 2024-04-12
Release/0.12.0

szczys:


sam-golioth:
Need to add RAK5010 to the list of boards in the README

CHANGELOG.md
```diff
@@ -4,6 +4,40 @@ All notable changes to this project will be documented in this file.
...
+
+- Measure code coverage in continuous integration tests
+- Add socket hello_nrf91_offload Zephyr example to demonstrate DTLS
+  sockets offload.
+- Zephyr: add rak5010_nrf52840 support to all examples
+- ESP-IDF: add fw_update example to demonstrate OTA firmware update
+
+### Changed:
+
+- Zephyr: example code now selects necessary Kconfig symbols explicitly
```
sam-golioth, line 25: This isn't accurate. The examples still use the defconfig files, but the defconfig is included explicitly by each example instead of implicitly included via the SDK. What this means is that _client_ applications can no longer use the defconfig files.
https://github.com/golioth/golioth-firmware-sdk/pull/420

CHANGELOG.md
```diff
@@ -4,6 +4,40 @@ All notable changes to this project will be documented in this file.
...
+### Removed:
+
+- ESP-IDF: Remove legacy hardware test which has been replaces by sample
+  tests and comprehensive tests that run on Hardware in the Loop (HIL).
+
+### Fixed:
+
+- Zephyr example readme now uses Golioth SDK version when using the
+  `west init` command
+- Various CI improvements
```
sam-golioth, line 39: We don't need to include CI changes here, as they don't directly impact users at all
```suggestion
```

CHANGELOG.md
```diff
@@ -4,6 +4,40 @@ All notable changes to this project will be documented in this file.
...
+
+### Known Issues:
+
+- Default Kconfig values may not propagate correctly if the Golioth SDK
+  is not listed first in the West manifest. To workaround this, ensure
+  that Golioth is listed first in your application's manifest.
+
+### Added:
+
+- Measure code coverage in continuous integration tests
```
sam-golioth, line 17: We don't need to include CI changes here, as they don't directly impact users at all
```suggestion
```

CHANGELOG.md
```diff
@@ -4,6 +4,40 @@ All notable changes to this project will be documented in this file.
...
+
+- Zephyr: example code now selects necessary Kconfig symbols explicitly
+  in the prj.conf file instead or relying on defconfig files (this
+  addresses the item mentioned in known issues).
+- Update libcoap version to v4.3.4a
+
+### Removed:
+
+- ESP-IDF: Remove legacy hardware test which has been replaces by sample
+  tests and comprehensive tests that run on Hardware in the Loop (HIL).
```
sam-golioth, line 33: We don't need to include CI changes here, as they don't directly impact users at all
```suggestion
```

CHANGELOG.md
```diff
@@ -4,6 +4,40 @@ All notable changes to this project will be documented in this file.
...
+
+### Fixed:
+
+- Zephyr example readme now uses Golioth SDK version when using the
```
sam-golioth, line 37:
```suggestion
- Documentation updates
```

CHANGELOG.md
```diff
@@ -4,6 +4,40 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.12.0] - 2024-04-15
+
+### Known Issues:
+
+- Default Kconfig values may not propagate correctly if the Golioth SDK
+  is not listed first in the West manifest. To workaround this, ensure
+  that Golioth is listed first in your application's manifest.
```
sam-golioth, line 13: This isn't an issue anymore. It's no longer relevant because client applications no longer have access to the default kconfig values used by the examples.
```suggestion
```

We should add a breaking change section describing this. Something like:
> Previously, it was possible to implicitly include default Kconfig settings from the Zephyr examples into client applications. This was unintentional and has been fixed. As a result, if your application was relying on those defaults, you'll now need to set those Kconfig values explicitly in your application.

CHANGELOG.md
```diff
@@ -4,6 +4,40 @@ All notable changes to this project will be documented in this file.
...
+
+### Added:
+
+- Measure code coverage in continuous integration tests
+- Add socket hello_nrf91_offload Zephyr example to demonstrate DTLS
+  sockets offload.
+- Zephyr: add rak5010_nrf52840 support to all examples
+- ESP-IDF: add fw_update example to demonstrate OTA firmware update
```
sam-golioth, line 21: We also added a Settings example

CHANGELOG.md
```diff
@@ -4,6 +4,40 @@ All notable changes to this project will be documented in this file.
...
+
+### Changed:
```
sam-golioth, line 23: We should also include the change to allow setting runtime credentials in the examples without a reboot.

examples/zephyr/certificate_provisioning/boards/nrf52840dk_nrf52840.conf
```diff
@@ -13,3 +13,6 @@ CONFIG_MCUMGR_GRP_FS_FILE_ACCESS_HOOK=y
 
 # Allow reboot from shell after loading certificates
 CONFIG_REBOOT=y
+
+CONFIG_NET_BUF_DATA_SIZE=256
```
sam-golioth, line 17: For the fw_update sample we set `CONFIG_NET_BUF_TX_COUNT=20` instead of changing `CONFIG_NET_BUF_DATA_SIZE`. I don't really have an opinion as to which is the better one to change, but would like to keep the examples consistent if possible. Have/can you try with `CONFIG_NET_BUF_TX_COUNT=20`?

szczys: This symbol is already resolves to 32 by for this application:
```
CONFIG_NET_BUF_TX_COUNT=32
```

I found that the change to NET_BUF_DATA_SIZE was necessary to get the handshake to complete successfully.

sam-golioth: Cool, let's do it.

CHANGELOG.md
szczys: I've updated this entry as follows:

```
- Zephyr: example code now explicitly sources Kconfig.defconfig file to 
  select necessary Kconfig symbols from the example common files.
```

CHANGELOG.md
```diff
@@ -4,6 +4,38 @@ All notable changes to this project will be documented in this file.
...
+  sockets offload.
+- Zephyr: add rak5010_nrf52840 support to all examples
+- ESP-IDF: add fw_update example to demonstrate OTA firmware update
+- ESP-IDF: add settings example to demonstrate device settings service
+
+### Changed:
+
+- Zephyr: example code now explicitly sources Kconfig.defconfig file to
+  select necessary Kconfig symbols from the example common files.
+- Zephyr: update runtime settings to take effect without needing a
```
sam-golioth, line 29:
```suggestion
- Zephyr: update runtime settings in examples to take effect without needing a
```

sam-golioth (approved):
1 small suggestion, looks good!

PR #445, opened 2024-04-16
coap_client: zephyr: handle connection stop

mniestroj:
Monitor `client->run_sem` semaphore, similar to how libcoap implementation does.

### DIFF
--- a/tests/hil/tests/connection/test.c
+++ b/tests/hil/tests/connection/test.c
@@ -11,6 +11,14 @@ void hil_test_entry(const struct golioth_client_config *config)
 
     while (1)
     {
-        golioth_sys_msleep(5000);
+        golioth_sys_msleep(10 * 1000);
+
+        GLTH_LOGI(TAG, "Stopping client");
+        golioth_client_stop(client);
+
+        golioth_sys_msleep(10 * 1000);
+
+        GLTH_LOGI(TAG, "Starting client");
+        golioth_client_start(client);
     }
 }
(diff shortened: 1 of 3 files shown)


tests/hil/tests/connection/test.c
```diff
@@ -1,16 +1,55 @@
...
 
     while (1)
     {
-        golioth_sys_msleep(5000);
+        golioth_sys_sem_take(sleep_temporarily_sem, GOLIOTH_SYS_WAIT_FOREVER);
+
+        GLTH_LOGI(TAG, "Stopping client");
+        golioth_client_stop(client);
+
+        golioth_sys_msleep(sleep_temporarily_msec);
```
sam-golioth, line 50: `sleep_temporarily_msec` could be uninitialized here if the RPC hasn't run yet, right?

tests/hil/tests/connection/test_connection.py
```diff
@@ -1,11 +1,28 @@
...
 
     # Confirm connection to Golioth
     assert None != board.wait_for_regex_in_line('Golioth CoAP client connected', timeout_s=120)
+
+
+async def test_stop_temporarily(board, device):
+    await trio.sleep(1)
+
+    try:
+        await device.rpc.stop_temporarily(10)
```
sam-golioth, line 22: Do we need the RPC if we have the start/stop in a loop? If we can manage it, would be nice to keep this test focused just on connections for isolation of testing.

mniestroj: Initially I wanted to control that from outside. Unfortunately we don't have generic (platform agnostic) way of running commands over serial (shell), so I went with RPC instead. But overall it is better to keep tested functionalities separate and do not use RPC if not needed.

Followed your suggestion about periodic disconnects and seems to work well for this testcase.

tests/hil/tests/connection/test.c
mniestroj: We would not reach this line, since we waited for semaphore that was released in RPC. So `sleep_temporarily_msec` was guaranteed to be set at this point. Anyway, this is outdated without RPC use.

PR #446, opened 2024-04-16
fw update: retry firmware observation

szczys:
If registering the firmware rollout observation fails, retry up to 3 times.

### DIFF
--- a/src/fw_update.c
+++ b/src/fw_update.c
@@ -17,6 +17,9 @@
 
 LOG_TAG_DEFINE(golioth_fw_update);
 
+#define FW_OBSERVATION_RETRY_COUNT 3
+#define FW_OBSERVATION_RETRY_TIMEOUT_MS 5000
+
 static struct golioth_client *_client;
 static golioth_sys_sem_t _manifest_rcvd;
 static struct golioth_ota_manifest _ota_manifest;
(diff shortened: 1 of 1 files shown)


src/fw_update.c
```diff
@@ -122,6 +122,45 @@ static bool manifest_version_is_different(const struct golioth_ota_manifest *man
...
+            break;
+        }
+        else if (retry_count == 1)
+        {
+            GLTH_LOGE(TAG, "Failed to observe firmware manifest: %d", status);
+        }
+        else
+        {
+            GLTH_LOGW(TAG, "Failed to observe manifest, will retry. %d", status);
+            golioth_sys_msleep(2000);
```
sam-golioth, line 158: Let's move the magic numbers (delay and retry count) to #defines at the top of the file. I don't think we need to make them kconfigurable just yet.

sam-golioth, line 158: Let's make this delay longer. It's not so important if we observe immediately after a reboot, but is important that we do eventually observe. Maybe 5 seconds?

src/fw_update.c
```diff
@@ -169,15 +208,7 @@ static void fw_update_thread(void *arg)
...
-    golioth_fw_update_report_state_sync(_client,
-                                        GOLIOTH_OTA_STATE_IDLE,
-                                        GOLIOTH_OTA_REASON_READY,
-                                        _config.fw_package_name,
-                                        _config.current_version,
-                                        NULL,
-                                        GOLIOTH_SYS_WAIT_FOREVER);
-
-    golioth_ota_observe_manifest_async(_client, on_ota_manifest, NULL);
+    fw_report_and_observe();
```
sam-golioth, line 211: :clap: for readability!

sam-golioth (approved):
Thanks for the quick turnaround!

PR #448, opened 2024-04-17
Blockwise Downloads Module with golioth_blockwise_get API

rashmivaidya72:


src/golioth_blockwise.h
```diff
@@ -0,0 +1,41 @@
...
+                    bool is_last,
+                    void *callback_arg);
+
+struct blockwise_transfer
+{
+    bool is_last;
+    enum golioth_status status;
+    int retry_count;
+    uint32_t offset;
+    size_t block_size;
```
sam-golioth, line 26: I'm wondering if some of these fields, like `block_size`, `status`, `is_last` need to be in the `ctx` struct or if we can carry them in local variables or function return values and arguments.

src/golioth_blockwise.c
```diff
@@ -0,0 +1,134 @@
...
+{
+    //assert valid values of arg, payload size and block_buffer
+    assert(arg);
+    assert(payload_size <= CONFIG_GOLIOTH_BLOCKWISE_DOWNLOAD_BUFF_SIZE);
+    struct blockwise_transfer *ctx = (struct blockwise_transfer *) arg;
+    assert(ctx->block_buffer);
+
+    //copy blockwise download values returned by COAP client into ctx
+    ctx->is_last = is_last;
+    memcpy(ctx->block_buffer , payload, payload_size);
```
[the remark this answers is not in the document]
rashmivaidya72: Yes, that's the reason I added the memcpy.

include/golioth/golioth_status.h
```diff
@@ -21,7 +21,11 @@
     STATUS(GOLIOTH_ERR_NOT_ALLOWED)         \
     STATUS(GOLIOTH_ERR_INVALID_STATE)       \
     STATUS(GOLIOTH_ERR_NO_MORE_DATA)        \
-    STATUS(GOLIOTH_ERR_NACK)
+    STATUS(GOLIOTH_ERR_NACK)                \
+    STATUS(GOLIOTH_ERR_INVALID_BLOCK_SIZE)  \
+    STATUS(GOLIOTH_ERR_MISMATCHED_OFFSET)   \
+    STATUS(GOLIOTH_ERR_NETWORK_TIMEOUT)     \
+    STATUS(GOLIOTH_ERR_CLOUD_SIDE_ERROR)
```
mniestroj, line 28: Introduced enums are not used anywhere.

src/Kconfig
```diff
@@ -62,6 +62,13 @@ config GOLIOTH_MAX_NUM_OBSERVATIONS
     help
         The maximum number of CoAP paths which can be simultaneously observed.
 
+config GOLIOTH_BLOCKWISE_DOWNLOAD_BUFF_SIZE
```
mniestroj, line 65: Please rename `s/BUFF/BUFFER/`. I think it will be more readable that way, since either `buf` or `buffer` are used as a name of a buffer.

src/coap_blockwise.c
```diff
@@ -0,0 +1,155 @@
...
+                                          write_block_cb cb,
+                                          void *callback_arg)
+{
+    enum golioth_status status = GOLIOTH_OK;
+    assert(client);
+    assert(path);
+    assert(cb);
+    struct blockwise_transfer *ctx = (struct blockwise_transfer *)malloc(
+        sizeof(struct blockwise_transfer));
+    uint8_t *data_buff = (uint8_t *)malloc(
```
mniestroj, line 138: There is no reason to cast explicitly to `(uint8_t *)`, just use implicit cast.

src/coap_blockwise.c
```diff
@@ -0,0 +1,155 @@
...
+        CONFIG_GOLIOTH_BLOCKWISE_DOWNLOAD_BUFF_SIZE * sizeof(uint8_t));
```
mniestroj, line 139: We can rather drop ` * sizeof(uint8_t)`, since it is single byte anyway.

mniestroj (commented):
I'd like to see how this replaces existing functionalities (e.g. part of the internal implementation of OTA) or how it extends functionalities that we provide already (by adding a sample and/or test usage).

include/golioth/golioth_status.h
rashmivaidya72: Right, I'll remove these for now and add them when I work on handling the error cases.

src/coap_blockwise.c
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// Function to download a single block
+static enum golioth_status download_single_block(struct golioth_client *client,
+                                                 struct blockwise_transfer *ctx)
+{
+    return golioth_coap_client_get_block(client,
+                                         ctx->path_prefix,
+                                         ctx->path,
+                                         GOLIOTH_CONTENT_TYPE_JSON,
```
sam-golioth, line 95: Just a note that this should probably be BINARY or OCTET_STREAM, but we don't support that in the Client yet, so it's ok to leave this as is for now (and matches the existing block download).

sam-golioth (approved):

PR #452, opened 2024-04-18
Obey limit when adding observations

szczys:
Check CONFIG_GOLIOTH_MAX_NUM_OBSERVATIONS before registering a new coap observation.

### DIFF
--- a/src/coap_client_libcoap.c
+++ b/src/coap_client_libcoap.c
@@ -513,7 +513,46 @@ static void golioth_coap_delete(golioth_coap_request_msg_t *req, coap_session_t 
     coap_send(session, req_pdu);
 }
 
-static void add_observation(golioth_coap_request_msg_t *req, struct golioth_client *client)
+static enum golioth_status golioth_coap_observe(golioth_coap_request_msg_t *req,
+                                                struct golioth_client *client,
+                                                coap_session_t *session)
+{
+    // GET with an OBSERVE option
+    coap_pdu_t *req_pdu = coap_new_pdu(COAP_MESSAGE_CON, COAP_REQUEST_CODE_GET, session);
+    if (!req_pdu)
+    {
+        GLTH_LOGE(TAG, "coap_new_pdu() get failed");
+        return GOLIOTH_ERR_MEM_ALLOC;
+    }
+
+    golioth_coap_add_token(req_pdu, req, session);
+
+    unsigned char optbuf[4] = {};
+    coap_add_option(req_pdu,
+                    COAP_OPTION_OBSERVE,
+                    coap_encode_var_safe(optbuf, sizeof(optbuf), COAP_OBSERVE_ESTABLISH),
+                    optbuf);
+
+    golioth_coap_add_path(req_pdu, req->path_prefix, req->path);
+    golioth_coap_add_accept(req_pdu, req->observe.content_type);
+
+    int err = coap_send(session, req_pdu);
+    if (err == COAP_INVALID_MID)
+    {
+        /* The coap_send() function returns the CoAP message ID on success or COAP_INVALID_MID on
+         * failure. */
+        GLTH_LOGE(TAG, "Unable to observe path %s, cannot send CoAP PDU", req->path);
+        return GOLIOTH_ERR_BAD_REQUEST;
+    }
+    else
+    {
+        return GOLIOTH_OK;
+    }
+}
+
+static enum golioth_status add_observation(golioth_coap_request_msg_t *req,
+                                           struct golioth_client *client,
+                                           coap_session_t *session)
 {
     // scan for available (not used) observation slot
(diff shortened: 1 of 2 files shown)


src/coap_client_zephyr.c
```diff
@@ -317,7 +317,29 @@ static int golioth_coap_get_block(golioth_coap_request_msg_t *req)
     return err;
 }
 
-static void add_observation(golioth_coap_request_msg_t *req, struct golioth_client *client)
+static int golioth_coap_observe(golioth_coap_request_msg_t *req, struct golioth_client *client)
+{
+    int err;
+
+    err = golioth_coap_req_cb(req->client,
```
sam-golioth, line 324:
```suggestion
    int err = golioth_coap_req_cb(req->client,
```

src/coap_client_zephyr.c
```diff
@@ -317,7 +317,29 @@ static int golioth_coap_get_block(golioth_coap_request_msg_t *req)
...
+                              golioth_coap_cb,
+                              req,
+                              GOLIOTH_COAP_REQ_OBSERVE);
+    if (err)
+    {
+        LOG_ERR("Failed to schedule CoAP OBSERVE: %d", err);
+        return err;
+    }
+
+    return 0;
```
sam-golioth, line 339:
```suggestion
    }

    return err;
```

src/coap_client_zephyr.c
```diff
@@ -495,8 +504,13 @@ static enum golioth_status coap_io_loop_once(struct golioth_client *client)
             break;
         case GOLIOTH_COAP_REQUEST_OBSERVE:
             LOG_DBG("Handle OBSERVE %s", req->path);
-            err = golioth_coap_observe(req, client);
-            add_observation(req, client);
+            err = add_observation(req, client);
+            if (err == GOLIOTH_ERR_QUEUE_FULL)
+            {
+                /* Observations are full, free the req but don't treat as a coap error */
+                err = GOLIOTH_OK;
```
sam-golioth, line 511: Why are we setting this to `GOLIOTH_OK`?

src/coap_client_libcoap.c
```diff
@@ -507,31 +507,6 @@ static void golioth_coap_delete(golioth_coap_request_msg_t *req, coap_session_t
...
-    {
-        GLTH_LOGE(TAG, "Unable to observe path %s, no slots available", req->path);
-        return;
-    }
-
-    obs_info->in_use = true;
-    memcpy(&obs_info->req, req, sizeof(obs_info->req));
-}
-
 static void golioth_coap_observe(golioth_coap_request_msg_t *req,
```
sam-golioth, line 510: Can we make similar changes as in the Zephyr implementation to bubble up errors?

szczys:
@sam-golioth I discovered an issue on when first creating an observation and added a commit to solve.

src/coap_client_zephyr.c
```diff
@@ -391,15 +391,19 @@ static int add_observation(golioth_coap_request_msg_t *req, struct golioth_clien
         return GOLIOTH_ERR_QUEUE_FULL;
     }
 
-    int err = golioth_coap_observe(req, client);
+    /* Store request message in Golioth client */
+    memcpy(&obs_info->req, req, sizeof(obs_info->req));
```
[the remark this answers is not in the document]
szczys: Interesting, we should discuss the details. I think it would involve changing the program flow of `coap_io_lop_once` which is where the request is allocated and populated from the mailbox.

https://github.com/golioth/golioth-firmware-sdk/blob/f5202d8688c083f45c22680fdd35dbf9aa4ed86d/src/coap_client_zephyr.c#L414-L426

src/coap_client_libcoap.c
```diff
@@ -531,37 +559,20 @@ static void add_observation(golioth_coap_request_msg_t *req, struct golioth_clie
...
-
-static void golioth_coap_observe(golioth_coap_request_msg_t *req,
-                                 struct golioth_client *client,
-                                 coap_session_t *session)
-{
-    // GET with an OBSERVE option
-    coap_pdu_t *req_pdu = coap_new_pdu(COAP_MESSAGE_CON, COAP_REQUEST_CODE_GET, session);
-    if (!req_pdu)
+    int err = golioth_coap_observe(req, client, session);
+    if (err == COAP_INVALID_MID)
```
sam-golioth, line 566: What about other errors?

szczys: This is the only error code that `send_coap()` returns ([libcoap docs](https://libcoap.net/doc/reference/4.3.4/man_coap_send.html#coap_send)). This issue here isn't the error testing, but where I'm doing it.

I've moved that check up into `golioth_coap_observe()`, left a comment about the error code, and changed the return type of the functions to be `enum golioth_status`.

sam-golioth (approved):

PR #457, opened 2024-04-22
Address build error in cpp and certificate_auth

rashmivaidya72:


rashmivaidya72:
Issue resolves after upgrading to esp-idf v5.2.1

PR #461, opened 2024-04-23
Retain original esp log formatting and introduce cloud logging option…

rashmivaidya72:
… default to 0


rashmivaidya72:
This commit is an improvement over the espidf log backend commit. The problem with the previous code was that it suddenly changed two things about the behavior of ESP_LOGX statements: 
1. it removed timestamp, module name, color formatting, log level from the terminal output 
2. it started sending logs to Golioth cloud. 

To rectify these migration problems, I refactored the code to maintain the original formatting, including color, timestamp, etc. I also added the config option for sending the logs to the cloud, which defaults to 0, so things will work as before by default.

These two cases work as expected:
Case 1:  CONFIG_GOLIOTH_ESPLOG_AUTO_LOG_TO_CLOUD=0 :
GLTH_LOGX sends logs to cloud
ESP_LOGX does not send logs to cloud

Case 2: CONFIG_GOLIOTH_ESPLOG_AUTO_LOG_TO_CLOUD=1 :
GLTH_LOGX sends logs to cloud
ESP_LOGX also sends logs to cloud

sam-golioth:
@rashmivaidya72 Now that v0.12.2 has been released, we can rebase this PR. We'll also need to add back in the original commit. For ease of reviewing (since we already reviewed the earlier commit), it would be best to have one commit that does the un-revert, and then these new changes in a separate commit.

PR #465, opened 2024-04-25
coap_client: zephyr: rework 'client stop'

mniestroj:
First of all, drop `client->end_session` member. This reduces amount of
data stored in a shared instance. Keep the information about being 'in
session' indirectly, by being IN or OUT of the while loop.

This patch fixes error and timeout handling as well.

`client->run_sem` was checked at the end of I/O loop. This was performed only
when the I/O loop had reached that stage and calling `golioth_client_stop()`
was not resulting in I/O loop wakeup, or more specifically wakeup from
`zsock_poll()` invocation. As a result it took up to 9s (default value of
`CONFIG_GOLIOTH_COAP_CLIENT_PING_INTERVAL_SEC`) to process "stop" request
from the application, when there was no traffic with Golioth cloud.

Implement a libcoap and Zephyr specific `golioth_client_start()` and
`golioth_client_stop()` APIs. For libcoap leave it as it was, i.e. relying on
1s polling of I/O loop. For Zephyr use `FLAG_STOP_CLIENT` flag that was not
used so far (it was a leftover from Zephyr SDK repo). Set that flag and
wakeup I/O loop by writing to existing eventfd (`POLLFD_EVENT`).


src/coap_client_libcoap.c
```diff
@@ -1179,6 +1179,77 @@ struct golioth_client *golioth_client_create(const struct golioth_client_config
...
 }
 
+static void purge_request_mbox(golioth_mbox_t request_mbox)
+{
+    golioth_coap_request_msg_t request_msg = {};
+    size_t num_messages = golioth_mbox_num_messages(request_mbox);
+
+    for (size_t i = 0; i < num_messages; i++)
+    {
+        assert(golioth_mbox_recv(request_mbox, &request_msg, 0));
```
sam-golioth, line 1189: I know this is copied code, but we shouldn't be wrapping function calls in `assert()`. Otherwise we could drop this call completely if asserts are turned off.
```suggestion
        bool ok = golioth_mbox_recv(request_mbox, &request_msg, 0);
        assert(ok);
```

src/coap_client_zephyr.c
```diff
@@ -1320,3 +1302,76 @@ struct golioth_client *golioth_client_create(const struct golioth_client_config
...
 }
+
+static void purge_request_mbox(golioth_mbox_t request_mbox)
+{
+    golioth_coap_request_msg_t request_msg = {};
+    size_t num_messages = golioth_mbox_num_messages(request_mbox);
+
+    for (size_t i = 0; i < num_messages; i++)
+    {
+        assert(golioth_mbox_recv(request_mbox, &request_msg, 0));
```
sam-golioth, line 1313: Same note about wrapping function calls in `assert()`
```suggestion
        bool ok = golioth_mbox_recv(request_mbox, &request_msg, 0);
        assert(ok);
```

PR #467, opened 2024-04-25
zephyr: coap: skip free for observations

szczys:
Observations are stored in the golioth client struct so that the data may be used to reestablish observations upon reconnect. When the Golioth client is stopped, all requests in the client->coap_reqs linked list are freed. This causes the client address in the stored observation to change, leading to a crash when trying to reestablish. This commit skips freeing observations so the data remains intact. This memory should only be freed when cancelling an observation.

resolves https://github.com/golioth/firmware-issue-tracker/issues/557

### DIFF
--- a/tests/hil/tests/rpc/README.md
+++ b/tests/hil/tests/rpc/README.md
@@ -1,3 +1,7 @@
 # RPC Test
 
 This tests the RPC service on the device.
+
+See [the HIL docs for the Zephyr
+platform](../../platform/zephyr/README.md) for an example of running
+tests locally.
--- a/src/coap_client_zephyr.c
+++ b/src/coap_client_zephyr.c
@@ -217,6 +217,11 @@ static int golioth_coap_cb(struct golioth_req_rsp *rsp)
         }
 
         err = rsp->err;
+        if (req->type == GOLIOTH_COAP_REQUEST_OBSERVE)
+        {
+            // don't free observations so we can reestablish later
+            return err;
+        }
         goto free_req;
     }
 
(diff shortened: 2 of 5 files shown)


szczys:
> This memory should only be freed when cancelling an observation.

Actually, with more thought this memory should **never** be freed. It's part of the `client` struct. There's a `in_use` flag that can be set to `false` when an observation is cancelled so the memory become available for a future observation.

sam-golioth:
@szczys Can we add a HIL test for this?

szczys:
> @szczys Can we add a HIL test for this?

I added a commit the puts a loop around the reestablish RPC tests. This takes the test through the client stop/start cycle 3 times. I tested it on main and it fails, but passes on this branch as expected.

tests/hil/tests/rpc/README.md
```diff
@@ -1,3 +1,29 @@
 # RPC Test
 
 This tests the RPC service on the device.
+
+## Run Zephyr test locally
```
sam-golioth, line 5: What do you think about putting this at `tests/hil/platform/zephyr/README.md`? We already have build instructions there. We can link to that doc from here, but I think it will be easier to maintain if we have instructions centrally located.

sam-golioth (approved):

tests/hil/tests/rpc/README.md
szczys: Sounds good. I've made the update and linked to that central README from the test readmes. I didn't link from the connection readme since that already has simplified syntax to run that test.

PR #468, opened 2024-04-29
zephyr: coap: add cancel observations

szczys:
Add ability to remove observations from the client coap_reqs lists, and mark the observation slots as not in_use. The cancellation process sends an "eager release" coap packet to the server indicating the observation has been  cancelled.

This searches the coap_reqs linked list by matching the request message pointer which is why the `coap_client.h` header was included. I think a future improvement to how observations are handled should be storing crucial observation values instead of the the request message. If we store the coap token as one of those values, it can be used to search in the coap_reqs.

* This should be merged after #467
* This should be merged after #452
* resolves https://github.com/golioth/firmware-issue-tracker/issues/519

### DIFF
--- a/src/coap_client.c
+++ b/src/coap_client.c
@@ -124,6 +124,12 @@ static enum golioth_status golioth_coap_client_set_internal(struct golioth_clien
         return GOLIOTH_ERR_INVALID_STATE;
     }
 
+    if (strlen(path) > sizeof(request_msg.path) - 1)
+    {
+        GLTH_LOGE(TAG, "Path too long: %d > %d", strlen(path), sizeof(request_msg.path) - 1);
+        return GOLIOTH_ERR_INVALID_FORMAT;
+    }
+
     if (payload_size > 0)
     {
         // We will allocate memory and copy the payload
@@ -150,7 +156,9 @@ static enum golioth_status golioth_coap_client_set_internal(struct golioth_clien
     request_msg.type = type;
     request_msg.path_prefix = path_prefix;
     request_msg.ageout_ms = ageout_ms;
+
     strncpy(request_msg.path, path, sizeof(request_msg.path) - 1);
+
     if (is_synchronous)
     {
         // Created here, deleted by coap thread (or here if fail to enqueue
@@ -316,7 +324,14 @@ enum golioth_status golioth_coap_client_delete(struct golioth_client *client,
             },
         .ageout_ms = ageout_ms,
     };
+
+    if (strlen(path) > sizeof(request_msg.path) - 1)
+    {
+        GLTH_LOGE(TAG, "Path too long: %d > %d", strlen(path), sizeof(request_msg.path) - 1);
+        return GOLIOTH_ERR_INVALID_FORMAT;
+    }
     strncpy(request_msg.path, path, sizeof(request_msg.path) - 1);
+
     enum golioth_status status = GOLIOTH_OK;
 
     if (is_synchronous)
@@ -395,7 +410,14 @@ static enum golioth_status golioth_coap_client_get_internal(struct golioth_clien
     enum golioth_status status = GOLIOTH_OK;
     request_msg.type = type;
     request_msg.path_prefix = path_prefix;
+
+    if (strlen(path) > sizeof(request_msg.path) - 1)
+    {
+        GLTH_LOGE(TAG, "Path too long: %d > %d", strlen(path), sizeof(request_msg.path) - 1);
+        return GOLIOTH_ERR_INVALID_FORMAT;
+    }
     strncpy(request_msg.path, path, sizeof(request_msg.path) - 1);
+
     if (is_synchronous)
     {
         // Created here, deleted by coap thread (or here if fail to enqueue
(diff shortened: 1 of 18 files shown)


src/coap_client_zephyr.c
```diff
@@ -407,6 +408,63 @@ static int add_observation(golioth_coap_request_msg_t *req, struct golioth_clien
...
+
+    err = coap_append_option_int(&packet, COAP_OPTION_ACCEPT, req->observe.content_type);
+    if (err)
+    {
+        LOG_ERR("Unable add content format to packet");
+        return err;
+    }
+
+    err = golioth_send_coap(client, &packet);
+
```
sam-golioth, line 477: NIT extra blank line

src/coap_client_zephyr.c
```diff
@@ -407,6 +408,63 @@ static int add_observation(golioth_coap_request_msg_t *req, struct golioth_clien
...
+    }
+
+    err = coap_packet_append_uri_path_from_pathv(&packet, pathv);
+    if (err)
+    {
+        LOG_ERR("Unable add path option");
+        return err;
+    }
+
+    err = coap_append_option_int(&packet, COAP_OPTION_ACCEPT, req->observe.content_type);
```
sam-golioth, line 469: Do we need this option? Since there aren't any payloads involved.

src/coap_client_zephyr.c
```diff
@@ -408,6 +408,25 @@ static int add_observation(golioth_coap_request_msg_t *req, struct golioth_clien
     return 0;
 }
 
+void golioth_cancel_all_observations(struct golioth_client *client)
+{
+    int err;
+    golioth_coap_observe_info_t *obs_info = NULL;
```
sam-golioth, line 414: These variables can both be declared at first use.

src/zephyr_coap_req.c
```diff
@@ -761,6 +762,59 @@ static void golioth_coap_reqs_cancel_all_with_reason(struct golioth_client *clie
...
+            if (coap_content_format < 0)
+            {
+                LOG_ERR("Unable to get coap content format from request: %d", coap_content_format);
+                return coap_content_format;
+            }
+
+            golioth_coap_request_msg_t *req_msg = req->user_data;
+
+            golioth_coap_req_cancel(req);
+            err = golioth_coap_client_observe_release_async(client,
```
sam-golioth, line 795: This feels a little wonky. `golioth_coap_request_msg_t` can be used for all requests, not just observations, right? So it's non-obvious looking at this function signature or the function comment that we will attempt to release an observation unconditionally while cancelling the request.

sam-golioth (commented):
Great work cutting through the cruft here @szczys!

The way zephyr_coap_req is interacting with coap_client_zephyr feels a little weird - can we see if there's a way to simplify that?

src/coap_client.c
```diff
@@ -486,6 +487,55 @@ enum golioth_status golioth_coap_client_observe_async(struct golioth_client *cli
...
+        .type = GOLIOTH_COAP_REQUEST_OBSERVE_RELEASE,
+        .path_prefix = path_prefix,
+        .ageout_ms = GOLIOTH_SYS_WAIT_FOREVER,
+        .observe =
+            {
+                .content_type = content_type,
+                .arg = arg,
+            },
+    };
+    strncpy(request_msg.path, path, sizeof(request_msg.path) - 1);
```
mniestroj, line 523: Would be good to check whether `path` fits into `request_msg.path`.

src/coap_client.c
```diff
@@ -486,6 +487,55 @@ enum golioth_status golioth_coap_client_observe_async(struct golioth_client *cli
...
+    {
+        GLTH_LOGW(TAG,
+                  "Client not running, dropping observe release request for path %s%s",
+                  path_prefix,
+                  path);
+        return GOLIOTH_ERR_INVALID_STATE;
+    }
+
+    golioth_coap_request_msg_t request_msg = {
+        .type = GOLIOTH_COAP_REQUEST_OBSERVE_RELEASE,
```
mniestroj, line 514: That is part of the commit before introducing `GOLIOTH_COAP_REQUEST_OBSERVE_RELEASE`.

src/coap_client.c
szczys: Is there a good way to get the length of this? It's a `const char *path` being passed in and I don't think I can safely call str(len) on that.

mniestroj: You should be able to use `strlen()` without any issues. If the length won't fit into `request_msg.path` then we should return some meaningful error code.

sam-golioth (changes requested):

src/coap_client.h
```diff
@@ -195,6 +196,15 @@ enum golioth_status golioth_coap_client_observe_async(struct golioth_client *cli
                                                       golioth_get_cb_fn callback,
                                                       void *callback_arg);
 
+enum golioth_status golioth_coap_client_observe_release_async(
```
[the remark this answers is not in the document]
szczys: Okay, I made a new commit to change the existing function name and updated the commits with the new functions to remove `_async` from the names.

sam-golioth (approved):
Great work on this one @szczys! Quite the beast to wrangle.

PR #469, opened 2024-04-29
examples: zephyr: Add CBOR examples to lightdb_stream example app

cdwilson:
Update the `lightdb_stream` example app to include sync & async examples showing how to send CBOR-encoded sensor data to LightDB Stream.

If you don't want the last commit (dc055c6464d0c34445f1a53275033b6e1dea2ca6) just LMK and I can drop it (was just trying to consolidate duplicate code)

### DIFF
--- a/examples/zephyr/lightdb_stream/pytest/test_sample.py
+++ b/examples/zephyr/lightdb_stream/pytest/test_sample.py
@@ -34,5 +34,5 @@ async def test_lightdb_stream(shell, device, wifi_ssid, wifi_psk):
             LOGGER.info("ts: {0}, temp: {1}".format(value['timestamp'], value['data']['sensor']['temp']))
             assert (value["data"]["sensor"]["temp"] == temp)
             temp += 0.5
-            if temp == 21.5:
+            if temp == 22.0:
                 break
--- a/examples/zephyr/lightdb_stream/src/main.c
+++ b/examples/zephyr/lightdb_stream/src/main.c
@@ -13,6 +13,7 @@ LOG_MODULE_REGISTER(stream, LOG_LEVEL_DBG);
 #include <samples/common/net_connect.h>
 #include <stdlib.h>
 #include <string.h>
+#include <zcbor_encode.h>
 #include <zephyr/drivers/sensor.h>
 #include <zephyr/kernel.h>
 
@@ -102,7 +103,7 @@ static void temperature_push_handler(struct golioth_client *client,
     return;
 }
 
-static void temperature_push_async(const struct sensor_value *temp)
+static void temperature_push_async_json(const struct sensor_value *temp)
 {
     char sbuf[sizeof("{\"temp\":-4294967295.123456}")];
     int err;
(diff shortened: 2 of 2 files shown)


sam-golioth:
@cdwilson Can we add tests for the new functionality? Should be as simple as extending the number of temperature readings we're checking (in `examples/zephyr/lightdb_stream/pytest/test_sample.y`) to ensure that we're covering the new functions.

examples/zephyr/lightdb_stream/pytest/test_sample.py
```diff
@@ -34,5 +34,5 @@ async def test_lightdb_stream(shell, device, wifi_ssid, wifi_psk):
             LOGGER.info("ts: {0}, temp: {1}".format(value['timestamp'], value['data']['sensor']['temp']))
             assert (value["data"]["sensor"]["temp"] == temp)
             temp += 0.5
-            if temp == 21.5:
+            if temp == 23.5:
```
cdwilson, line 37: > Can we add tests for the new functionality? Should be as simple as extending the number of temperature readings we're checking (in examples/zephyr/lightdb_stream/pytest/test_sample.y) to ensure that we're covering the new functions.

@sam-golioth is this what you were thinking?

sam-golioth: Yep! We only added two new set operations, so `22.5` should suffice.

cdwilson: Ahh, got it. I misread the test previously and thought it was checking each set operation twice.

Based on your comments, I believe that the intent is just to check that we get a single valid value for each set operation, right? If that's the case, couldn't the new value actually be `22.0`?

i.e.

1. JSON (sync): `20.0` <- initial value of `temp`
2. JSON (async): `20.5`
3. CBOR (sync): `21.0`
4. CBOR (async): `21.5`
5. (loop breaks at `22.0` but doesn't check this value)

sam-golioth: Ah yeah, that's right. We previously had a `lightdb_stream_set_int()` function that we were also testing, but it's been removed, probably forgot to update the test.

cdwilson: I just changed it to `22.0`. I think this PR should be ready for review now.

examples/zephyr/lightdb_stream/src/main.c
```diff
@@ -102,48 +103,117 @@ static void temperature_push_handler(struct golioth_client *client,
...
+                                       NULL);
+    }
+    else
+    {
+        err = golioth_stream_set_sync(client,
+                                      "sensor",
+                                      GOLIOTH_CONTENT_TYPE_JSON,
+                                      sbuf,
+                                      strlen(sbuf),
+                                      2);
```
cdwilson, line 130: @sam-golioth I ended up increasing this timeout from 1s to 2s, and it appears to have fixed [the lightdb_stream test that was failing yesterday due to a timeout](https://github.com/golioth/golioth-firmware-sdk/actions/runs/8913757172/job/24480299831). I've run the tests a couple times and haven't seen this specific failure again, although I haven't been able to get a 100% pass due to other tests randomly failing.

sam-golioth: That makes sense, the failure you saw was on the 9160dk and we've definitely seen cell network transactions take longer than 1 second in the past.

sam-golioth (approved):

PR #470, opened 2024-04-29
examples: zephyr: Fix get_temperature with temp0 DT alias

cdwilson:
When a `temp0` devicetree alias is added to the board overlay, the `get_temperature` function was looping through all sensor channels, always returning the result of reading the last one, even if it was not a supported channel.

This fixes the function so that it returns after the first successful read from a temperature channel.


sam-golioth (approved):
Nice find!

Let's hold off on merging this until after v0.12.2 is released.

PR #475, opened 2024-04-30
Blockwise uploads support for POST_BLOCK request type

rashmivaidya72:
NOTE: I have not tested blockwise uploads on Zephyr yet.
Edit: tested for Zephyr as well, works as expected.


src/coap_client.h
```diff
@@ -61,7 +75,7 @@ typedef enum
     GOLIOTH_COAP_REQUEST_GET,
     GOLIOTH_COAP_REQUEST_GET_BLOCK,
     GOLIOTH_COAP_REQUEST_POST,
-    // TODO - support for POST_BLOCK
```
sam-golioth, line 64: 👏

src/coap_client_libcoap.c
```diff
@@ -499,6 +535,43 @@ static void golioth_coap_post(golioth_coap_request_msg_t *req, coap_session_t *s
...
+    {
+        GLTH_LOGE(TAG, "coap_new_pdu() post failed");
+        return;
+    }
+
+    if (req->post_block.block_index == 0)
+    {
+        // Save this token for further blocks
+        golioth_coap_add_token(req_pdu, req, session);
+        memcpy(client->block_token, req->token, req->token_len);
```
sam-golioth, line 553: Is `client->block_token` also used for block downloads? Seems like this could get clobbered if we try to do more than one blockwise operation at a time.

If the answer is yes, let's file a ticket for allowing multiple blockwise transfers at the same time. No need to address in this PR.

sam-golioth (approved):

src/coap_client_libcoap.c
rashmivaidya72: Yes, it is used by both. Here's the new issue:
https://github.com/golioth/firmware-issue-tracker/issues/569

PR #479, opened 2024-05-02
Blockwise Uploads Module with golioth_blockwise_post API

rashmivaidya72:


src/coap_blockwise.c
```diff
@@ -22,13 +22,144 @@ struct blockwise_transfer
...
+// Function to upload a single block
+static enum golioth_status upload_single_block(struct golioth_client *client,
+                                               struct blockwise_transfer *ctx)
+{
+    assert(ctx->block_size > 0 && ctx->block_size <= CONFIG_GOLIOTH_BLOCKWISE_UPLOAD_BLOCK_SIZE);
+    return golioth_coap_client_set_block(client,
+                                         ctx->path_prefix,
+                                         ctx->path,
+                                         ctx->is_last,
+                                         GOLIOTH_CONTENT_TYPE_JSON,
```
szczys, line 87: I'm curious about what determines the type needed here?

src/coap_blockwise.c
```diff
@@ -22,13 +22,144 @@ struct blockwise_transfer
...
+{
+    enum golioth_status status = GOLIOTH_ERR_FAIL;
+    assert(client);
+    assert(path);
+    assert(cb);
+
+    struct blockwise_transfer *ctx = malloc(sizeof(struct blockwise_transfer));
+    uint8_t *data_buff = malloc(CONFIG_GOLIOTH_BLOCKWISE_UPLOAD_BLOCK_SIZE);
+    assert(data_buff);
+    assert(ctx);
```
szczys, line 141: Is it safe to use assert here rather than issuing an error message and returning? For instance, does this cause the program to crash if `malloc()` fails due to lack of memory?

src/coap_blockwise.c
```diff
@@ -22,13 +22,152 @@ struct blockwise_transfer
...
+// Function to call the application's read block callback after a successful
+// blockwise upload
+static void call_read_block_callback(struct blockwise_transfer *ctx)
+{
+    if (ctx->callback.read_cb(ctx->offset,
+                              ctx->block_buffer,
+                              &ctx->block_size,
+                              &ctx->is_last,
+                              ctx->callback_arg)
+        != GOLIOTH_OK)
```
sam-golioth, line 72: I think this would be easier to read.
```suggestion
    int err = ctx->callback.read_cb(ctx->offset,
                                    ctx->block_buffer,
                                    &ctx->block_size,
                                    &ctx->is_last,
                                    ctx->callback_arg);
    if (err != GOLIOTH_OK)
```

src/coap_blockwise.c
sam-golioth: This should probably be binary (which we don't support yet, but doing so is straightforward), or better yet, we should let the user set the content type.

src/coap_blockwise.c
```diff
@@ -22,13 +22,152 @@ struct blockwise_transfer
...
+    struct blockwise_transfer *ctx = malloc(sizeof(struct blockwise_transfer));
+    if (!ctx)
+    {
+        return GOLIOTH_ERR_MEM_ALLOC;
+    }
+
+    uint8_t *data_buff = malloc(CONFIG_GOLIOTH_BLOCKWISE_UPLOAD_BLOCK_SIZE);
+    if (!data_buff)
+    {
+        return GOLIOTH_ERR_MEM_ALLOC;
```
sam-golioth, line 148: We need to free `ctx` here before returning.

src/coap_blockwise.c
```diff
@@ -131,13 +270,24 @@ enum golioth_status golioth_blockwise_get(struct golioth_client *client,
...
+    if (!ctx)
+    {
+        return GOLIOTH_ERR_MEM_ALLOC;
+    }
+
     uint8_t *data_buff = malloc(CONFIG_GOLIOTH_BLOCKWISE_DOWNLOAD_BUFFER_SIZE);
-    assert(data_buff);
+    if (!data_buff)
+    {
+        return GOLIOTH_ERR_MEM_ALLOC;
```
sam-golioth, line 288: We need to free `ctx` here as well

sam-golioth (approved):
Great work @rashmivaidya72!

PR #480, opened 2024-05-03
settings: check for error when responding to server

szczys:
If a resonse to a settings update is not sent back to the server, the device settings status will not be updated. This commit reports failure to send the response as a log message.


sam-golioth (approved):
👏

PR #481, opened 2024-05-06
Add settings cancel API

szczys:
Add API to cancel settings observations and free the settings service context.

Resolves https://github.com/golioth/firmware-issue-tracker/issues/521

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -467,6 +467,20 @@ finish:
     return gsettings;
 }
 
+enum golioth_status golioth_settings_deinit(struct golioth_settings *settings)
+{
+
+    if (settings == NULL)
+    {
+        GLTH_LOGE(TAG, "Settings service handle must not be NULL");
+        return GOLIOTH_ERR_NULL;
+    }
+
+    golioth_coap_client_cancel_observations_by_prefix(settings->client, SETTINGS_PATH_PREFIX);
+    free(settings);
+    return GOLIOTH_OK;
+}
+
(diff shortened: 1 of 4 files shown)


src/settings.c
```diff
@@ -463,6 +463,21 @@ struct golioth_settings *golioth_settings_init(struct golioth_client *client)
     return gsettings;
 }
 
+enum golioth_status golioth_settings_deinit(struct golioth_client *client,
```
sam-golioth, line 466: Same comment as the RPC PR: We can extract `client` from `settings` so this function only needs to take one argument.

tests/hil/tests/settings/test.c
```diff
@@ -85,10 +98,49 @@ void hil_test_entry(const struct golioth_client_config *config)
...
+    cancel_all_sem = golioth_sys_sem_create(1, 0);
+
+    golioth_debug_set_cloud_log_enabled(false);
+
+    struct golioth_client *client = golioth_client_create(config);
+    golioth_client_register_event_callback(client, on_client_event, NULL);
+
+    golioth_sys_sem_take(connected_sem, GOLIOTH_SYS_WAIT_FOREVER);
+
+    perform_settings_registration(client);
```
sam-golioth, line 118: If we change this function to return `struct golioth_settings *` then we can keep `settings` as a local stack variable instead of moving it to a global.

sam-golioth (approved):

sam-golioth (approved):
Looks good!

PR #483, opened 2024-05-07
HIL: adjust OTA test timing

szczys:
Adjust OTA test timing to reduce tests that fail due to propagation delay.

https://github.com/golioth/firmware-issue-tracker/issues/568

### DIFF
--- a/tests/hil/tests/ota/test.c
+++ b/tests/hil/tests/ota/test.c
@@ -95,7 +95,7 @@ static void test_reason_and_state(void)
         {
             GLTH_LOGE(TAG, "Unable to report ota state: %d", status);
         }
-        golioth_sys_msleep(5000);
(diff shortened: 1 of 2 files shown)


sam-golioth:
Have you seen tests fail at the margins of these timeouts? Or is this more of a guess and check?

szczys:
More of a guess and check. I have not been able to observe a status update sent to the server fail to update when I check manually so I am guessing that the 1.5s timeout pytest uses from seeing the log message to running the API call is the culprit. I doubled that to 3 seconds and built in some space for that in the C code timing.

I'd prefer not to bank on these timings, but the alternative that comes to mind is introducing an RPC to tell the device to go to the next status. That doesn't seem like a good idea.

sam-golioth (approved):

mniestroj (approved):

PR #488, opened 2024-05-09
zephyr: coap: synthesize host IPv6 address from uri with Thread

MarkoPura:
Leverage the OpenThread DNS to synthesize the Golioth System Server IPv6 address from GOLIOTH_COAP_HOST_URI to avoid hardcoded IP address in applications. NAT64 prefix used by the Thred Border Router is set while synthesizing the address.


src/golioth_openthread.c
```diff
@@ -0,0 +1,77 @@
...
+
+#include <openthread/nat64.h>
+#include <openthread/dns_client.h>
+#include <zephyr/net/openthread.h>
+#include <openthread/error.h>
+
+static otDnsQueryConfig dns_query;
+static char *golioth_ipv6_addr;
+
+K_SEM_DEFINE(ot_dns_resolve, 0, 1);
```
mniestroj, line 18: It is better to not use any global variables like that.

Please create a helper structure that you will pass using `void *aContext` to `ot_dns_callback()`:
```
struct ot_dns_resolve_context {
    struct k_sem sem;
    char *ipv6;
    otDnsQueryConfig query_config; /* optional, if you need it in the callback */
};
```
then access those members in `ot_dns_callback`.

src/golioth_openthread.c
```diff
@@ -0,0 +1,77 @@
...
+
+/* Callback for NAT64 IPv6 translated Golioth System Server address from the DNS query response */
+static void ot_dns_callback(otError aError, const otDnsAddressResponse *aResponse, void *aContext)
+{
+    otIp6Address golioth_addr;
```
mniestroj, line 23:
```
struct ot_dns_resolve_context *context = aContext;
```

and instead of `golioth_ipv6_addr` you can use `context->ipv6`. Instead of `k_sem_give(&ot_dns_resolve)` use `k_sem_give(&context->sem)`.

src/golioth_openthread.c
```diff
@@ -0,0 +1,77 @@
...
+
+    if (aError != OT_ERROR_NONE)
+    {
+        LOG_ERR("Golioth System Server DNS resolving error: %d", aError);
+        return;
+    }
```
mniestroj, line 29: Make sure to pass the error code to the caller (`synthesize_ipv6_address`) so it can continue execution (instead of indefinitely waiting for semaphore) and return this error code. You can use the same `struct ot_dns_resolve_context` structure for that, with additonal member in there `int err;` or something.

src/golioth_openthread.c
```diff
@@ -0,0 +1,77 @@
...
+					                   ot_dns_callback,
+					                   ot_context,
+					                   &dns_query);
+    if (err != OT_ERROR_NONE)
+    {
+        LOG_ERR("Golioth System Server address resolution DNS query error: %d", err);
+        return err;
+    }
+
+    k_sem_take(&ot_dns_resolve, K_FOREVER);
```
mniestroj, line 74: Using `k_sem` means that this code will work only under Zephyr. If that is the case for now and foreseeable future, then please move this file to Zephyr specific directory, instead of `src/`.

src/golioth_openthread.c
```diff
@@ -0,0 +1,77 @@
...
+    if (err != OT_ERROR_NONE)
+    {
+        LOG_ERR("Synthesize DNS server IPv6 address error: %d", err);
+        return err;
+    }
+
+    err = otDnsClientResolveIp4Address(ot_context->instance,
+					                   hostname,
+					                   ot_dns_callback,
+					                   ot_context,
```
mniestroj, line 66: You don't use `ot_context` in the callback, so please pass pointer to `struct ot_dns_resolve_context` here, so it will be possible to access it using `void *aContext` in the callback.

src/coap_client_zephyr.c
```diff
@@ -761,6 +762,18 @@ static int golioth_connect(struct golioth_client *client)
...
+    char ipv6_addr[32];
+    err = synthesize_ipv6_address(host, ipv6_addr);
+    if (err)
+    {
+        LOG_ERR("Failed to synthesize Golioth Server IPv6 address: %d", err);
+        return err;
+    }
+
+    host = ipv6_addr;
+#endif
```
mniestroj, line 775: If possible use:
```
if (IS_ENABLED(CONFIG_NET_L2_OPENTHREAD) {
    err = synthesize_ipv6_address(host, ipv6_addr);
    if (err)
    {
        LOG_ERR("Failed to synthesize Golioth Server IPv6 address: %d", err);
        return err;
    }

    host = ipv6_addr;
}
```

so that code will always compile and we will detect build errors quickly during any other regular builds.

mniestroj (dismissed):

src/coap_client_zephyr.c
```diff
@@ -740,6 +741,7 @@ static int golioth_connect(struct golioth_client *client)
 {
     char uri[] = CONFIG_GOLIOTH_COAP_HOST_URI;
     char *host = &uri[sizeof("coaps://") - 1];
+    char ipv6_addr[32];
```
mniestroj, line 744: Is 32 enough here? How about using `OT_IP6_ADDRESS_STRING_SIZE` (40) instead?

port/zephyr/include/golioth_openthread.h
```diff
@@ -0,0 +1,24 @@
...
+ * Get the IPv6 address of Golioth Server to avoid hardcoding it in applications.
+ * NAT64 prefix used by the Thread Border Router is set while synthesizing the address.
+ *
+ * @param[in] hostname Pointer to the host name for which to querry the address
+ * @param[out] ipv6_addr_buffer Buffer to char array to output the synthesized IPv6 address
+ *
+ * @retval 0 On success
+ * @retval <0 On failure
+ */
+int synthesize_ipv6_address(char *hostname, char *ipv6_addr_buffer);
```
mniestroj, line 22: We should probably namespace this function somehow. E.g. with `golioth_ot_` prefix. @sam-golioth any thoughts?

mniestroj (commented):
Looks good! The missing thing is namespacing function in one of my comments, otherwise good to merge.

src/coap_client_zephyr.c
MarkoPura: Good call! I used `40` as the array size in Zephyr SDK. I will change it to `OT_IP6_ADDRESS_STRING_SIZE`.

sam-golioth (approved):

mniestroj (approved):

PR #489, opened 2024-05-09
Error handling with retries in blockwise upload

rashmivaidya72:


src/coap_blockwise.c
```diff
@@ -63,65 +69,113 @@ static void on_block_sent(struct golioth_client *client,
...
 // Function to handle blockwise upload errors
-static enum golioth_status handle_upload_error()
+// This function retries for all errors like invalid state, failed mem alloc,
+// queue full, disconnects, etc returned by golioth_coap_client_set_block
+static enum golioth_status handle_upload_error(struct golioth_client *client,
+                                               struct blockwise_transfer *ctx)
 {
-    // TODO: handle errors like disconnects etc
-    return GOLIOTH_OK;
+    while (1)
```
sam-golioth, line 139: We should check (and decrement) the retry count here in the `while()`. I also don't think we need to keep `retry_count` in the context struct.

src/coap_blockwise.c
```diff
@@ -134,10 +188,16 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
                                            read_block_cb cb,
                                            void *callback_arg)
 {
-    enum golioth_status status = GOLIOTH_ERR_FAIL;
+    enum golioth_status status;
```
sam-golioth, line 191: This can be declared closer to use.

src/coap_blockwise.c
```diff
@@ -63,65 +69,113 @@ static void on_block_sent(struct golioth_client *client,
...
-    enum golioth_status status = GOLIOTH_ERR_FAIL;
-    ctx->block_size = 0;
-    call_read_block_callback(ctx);
-    if (ctx->block_size)
+    enum golioth_status status = call_read_block_callback(ctx);
+    if (status == GOLIOTH_OK)
     {
-        status = upload_single_block(client, ctx);
+        upload_single_block(client, ctx);
         if (ctx->status == GOLIOTH_OK)
```
sam-golioth, line 172: It feels confusing to be mixing storing the status in the context and keeping it local on the stack. I'm not sure that we need it in the context struct.

src/coap_blockwise.c
```diff
@@ -63,65 +69,113 @@ static void on_block_sent(struct golioth_client *client,
...
         {
             ctx->offset++;
         }
         else
         {
-            // TODO: handle_upload_error
-            status = handle_upload_error();
+            status = handle_upload_error(client, ctx);
```
sam-golioth, line 178: If we fall into the `else` clause we never increment the offset.

src/coap_blockwise.c
sam-golioth: I see. The status in the context struct is used to store the status that we get from the server response that is passed to us via a callback executed on the CoAP client thread. But because we are doing synchronous operations, the server response status is also returned to us directly, so we don't even actually need the callback.

On the other hand, the synchronous functions internal to the coap client create and destroy a semaphore for each transaction (i.e. for each block that we download). It would be more efficient to use asynchronous transfers and to make the operations blocking by waiting on a semaphore that is managed by the blockwise transfers module and created once per blockwise transfer.

PR #492, opened 2024-05-14
coap: add octet stream content type for binary data

szczys:
resolves https://github.com/golioth/firmware-issue-tracker/issues/502

### DIFF
--- a/src/coap_client_libcoap.c
+++ b/src/coap_client_libcoap.c
@@ -390,8 +390,9 @@ static uint32_t golioth_content_type_to_coap_type(enum golioth_content_type cont
             return COAP_MEDIATYPE_APPLICATION_JSON;
         case GOLIOTH_CONTENT_TYPE_CBOR:
             return COAP_MEDIATYPE_APPLICATION_CBOR;
+        case GOLIOTH_CONTENT_TYPE_OCTET_STREAM:
+            return COAP_MEDIATYPE_APPLICATION_OCTET_STREAM;
         default:
-            assert(0);
             return COAP_MEDIATYPE_APPLICATION_OCTET_STREAM;
     }
 }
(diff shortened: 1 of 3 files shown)


src/coap_client_libcoap.c
```diff
@@ -390,9 +390,10 @@ static uint32_t golioth_content_type_to_coap_type(enum golioth_content_type cont
             return COAP_MEDIATYPE_APPLICATION_JSON;
         case GOLIOTH_CONTENT_TYPE_CBOR:
             return COAP_MEDIATYPE_APPLICATION_CBOR;
+        case GOLIOTH_CONTENT_TYPE_OCTET_STREAM:
+            return COAP_MEDIATYPE_APPLICATION_OCTET_STREAM;
         default:
             assert(0);
-            return COAP_MEDIATYPE_APPLICATION_OCTET_STREAM;
```
hasheddan, line 395: I know that we don't expect to hit this case (hence the `assert(0)`), but I think we should keep the default return value as octet stream here in case the user is compiling with asserts disabled.

src/coap_client_libcoap.c
```diff
@@ -391,7 +391,6 @@ static uint32_t golioth_content_type_to_coap_type(enum golioth_content_type cont
         case GOLIOTH_CONTENT_TYPE_CBOR:
             return COAP_MEDIATYPE_APPLICATION_CBOR;
         default:
```
sam-golioth, line 393: We should also add an explicit case for `GOLIOTH_CONTENT_TYPE_OCTET_STREAM` here (it can fall through). That makes the code a little more maintainable so it's easier to see at a glance what the intention is and makes the code more searchable. In the future, when we have static analysis, it will also tell the analyzer that we intend for that case to use the default value.

sam-golioth (approved):

PR #493, opened 2024-05-15
lightdb: require content type when registering an observation

szczys:
Allow user to specify the LightDB State serialization format returned by the server when an observation is registerd. Prior to this change, JSON was being used by default while the server is also capable of returning CBOR.


tests/hil/tests/lightdb/test.c
```diff
@@ -395,7 +395,11 @@ static void test_lightdb_invalid(struct golioth_client *client)
 
 static void test_lightdb_observe_setup(struct golioth_client *client)
 {
-    golioth_lightdb_observe_async(client, "hil/lightdb/observed/int", int_cb, NULL);
+    golioth_lightdb_observe_async(client,
```
sam-golioth, line 398: Let's add a test for CBOR observations too

sam-golioth (approved):
Nice change!

tests/hil/tests/lightdb/test.c
szczys: I've added a commit to test both CBOR and JSON observation types.

sam-golioth (approved):

PR #494, opened 2024-05-17
kconfig: re-organize Kconfig symbols for menuconfig

szczys:
Break Kconfig symbols out into files that group them by type. Update the Zephyr Kconfig to include the various new Kconfig groups into menus for better presentation in menuconfig.

resolves https://github.com/golioth/firmware-issue-tracker/issues/571

### DIFF
--- a/src/Kconfig.coap
+++ b/src/Kconfig.coap
@@ -1,0 +1,84 @@
+#
+# Copyright (C) 2024 Golioth, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+#
+
+config GOLIOTH_COAP_HOST_URI
+    string "CoAP server URI"
+    default "coaps://coap.golioth.io"
+    help
+        The URI of the CoAP server
+
+config GOLIOTH_COAP_RESPONSE_TIMEOUT_S
+    int "CoAP response timeout"
+    default 10
+    help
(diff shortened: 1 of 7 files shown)


src/Kconfig.coap
```diff
@@ -0,0 +1,90 @@
...
+        Set to 0 to disable.
+
+config GOLIOTH_COAP_MAX_PATH_LEN
+    int "Golioth maximum CoAP path length"
+    default 39
+    help
+        Maximum length of a CoAP path (everything after
+        "coaps://coap.golioth.io/").
+
+config GOLIOTH_LIBCOAP_EXTRA_FDS_MAX
```
sam-golioth, line 86: Should we move this to somewhere in the `port/` folder? So it only shows up for ports that are using libcoap.

sam-golioth (commented):
This is awesome, thanks @szczys!

src/Kconfig.coap
szczys: I grepped the SDK and I don't see this symbol referenced anywhere else. I think it's a leftover from when we used libcoap with Zephyr so I removed it along with the symbol mentioned in the comment above. (This removal was done in a separate commit from the reorg work).

sam-golioth (approved):

PR #495, opened 2024-05-18
OTA refactoring to use blockwise download module

rashmivaidya72:
NOTE: Compilation of OTA hil tests fail because the function golioth_ota_get_block_sync() is no longer available. Need to update the test to use golioth_blockwise_get() function instead.


sam-golioth:
Superseded by #551

PR #496, opened 2024-05-21
hil: increase settings response buffer

szczys:
A new setting was added to the CI project which causes use to run out of space in the settings response buffer (used to report out-of-sunc settings). The result is that several settings test are failing on a no more core error:

```
<err> golioth_settings: Failed to send a response to server: -12
```

This commit doubles that buffer to avoid the error.


sam-golioth (changes requested):
I think we need a corresponding change for Linux in golioth_user_config.h

sam-golioth (approved):

PR #497, opened 2024-05-21
ci: cancel previous in progress runs

sam-golioth:


.github/workflows/hil_test_trigger.yml
```diff
@@ -8,6 +8,10 @@ on:
     # Run workflow at the start of every day (12 AM UTC)
     - cron: "0 0 * * *"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
mniestroj, line 13: Could you elaborate in commit message what it does, why we want this change, what it solves?

.github/workflows/hil_test_trigger.yml
```diff
@@ -8,6 +8,10 @@ on:
     # Run workflow at the start of every day (12 AM UTC)
     - cron: "0 0 * * *"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
szczys, line 12: I'm curious about what the second half of the conditional does. I see this in the github docs and it seems like it will cancel all HIL workworkflows, even ones unrelated to this PR:

> ```
> concurrency:
>   group: ${{ github.ref }}
>   cancel-in-progress: true
> ```
> 
> Note that in this example, without defining a particular concurrency group, GitHub Actions will cancel any in-progress run of the job or workflow.
>

.github/workflows/hil_test_trigger.yml
sam-golioth: Updated! Thanks for keeping me honest 🙂

mniestroj (approved):

PR #499, opened 2024-05-23
tests: hil: rpc: wait for RPC establishment before sending RPCs

sam-golioth:
In the observation_repeat_restart test, we were waiting for the device to establish a connection to Golioth before the test script sends an RPC to the device. This creates a small window between when the device establishes a connection and when it establishes an observation on the RPC path where attempts to send an RPC will fail. Instead, we should wait for the device to report that is successfully established an observation before the test script attempts to send any RPCs to the device. This matches the behavior in other parts of the test script.

### DIFF
--- a/tests/hil/tests/rpc/test.c
+++ b/tests/hil/tests/rpc/test.c
@@ -47,6 +47,8 @@ static enum golioth_rpc_status on_basic_return_type(zcbor_state_t *request_param
     // one of {int, string, float, bool}
     struct zcbor_string type;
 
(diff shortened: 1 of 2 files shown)


szczys (approved):

mniestroj (approved):

sam-golioth:
@szczys @mniestroj I added another commit with additional logging, please take another look

mniestroj:
> I added another commit with additional logging, please take another look

LGTM

PR #500, opened 2024-05-23
zephyr: coap: prevent resend backend logs

szczys:
When a network connection is unstable, coap packets may be resent. Logging this information is useful, but if backend logging is enabled, these get added to the network queue, further compounding the problem. Check to ensure backend logging is disabled before creating a log message about a coap request being resent.

### DIFF
--- a/src/zephyr_coap_req.c
+++ b/src/zephyr_coap_req.c
@@ -697,6 +697,8 @@ static int64_t golioth_coap_req_poll_prepare(struct golioth_coap_req *req, uint3
 
     if (send)
     {
+#ifndef CONFIG_LOG_BACKEND_GOLIOTH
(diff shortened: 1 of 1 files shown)


src/zephyr_coap_req.c
```diff
@@ -697,13 +697,18 @@ static int64_t golioth_coap_req_poll_prepare(struct golioth_coap_req *req, uint3
 
     if (send)
     {
+#ifndef CONFIG_LOG_BACKEND_GOLIOTH
```
sam-golioth, line 700: Is there a way we can log this to all backends except for Golioth, rather than not logging it at all?

sam-golioth:
@szczys Let's close this

PR #504, opened 2024-05-29
Update remaining LightDB Stream links

hasheddan:
Fixes up a few straggling LightDB Stream links.

Mirrors changes made in https://github.com/golioth/golioth-firmware-sdk/commit/900678a3f363055258c4946b3be9c5222dad5ea4

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ instructions to build and run the examples.
 The `golioth_basics` example (`examples/<platform>/golioth_basics`) is recommended
 as a starting point, to learn how to connect to Golioth and use services like
 [LightDB State](https://docs.golioth.io/cloud/services/lightdb),
-[Stream](https://docs.golioth.io/cloud/services/lightdb-stream),
+[Stream](https://docs.golioth.io/data-routing),
 [Logging](https://docs.golioth.io/cloud/services/logging),
 and [OTA](https://docs.golioth.io/cloud/services/ota).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -55,7 +55,7 @@ instructions to build and run the examples.
 The `golioth_basics` example (`examples/<platform>/golioth_basics`) is recommended
 as a starting point, to learn how to connect to Golioth and use services like
 [LightDB State](https://docs.golioth.io/cloud/services/lightdb),
-[Stream](https://docs.golioth.io/cloud/services/lightdb-stream),
+[Stream](https://docs.golioth.io/data-routing),
 [Logging](https://docs.golioth.io/cloud/services/logging),
 and [OTA](https://docs.golioth.io/cloud/services/ota).
```
hasheddan, line 60: note: it appears that none of these redirects have been working as expected (i.e. all routed to application services), but that should be fixed by https://github.com/golioth/docs/pull/393

sam-golioth: Should we just go ahead and update these to point to the right links?

hasheddan: @sam-golioth I think the intention is for these `/cloud/services/*` links to effectively act as short links to wherever we want them to go now and in the future. The `LightDB Stream` -> `Stream` example is a bit of an anomaly as LightDB Stream is still a cloud service, but its not what we are actually sending directly to from the firmware SDK anymore, so we couldn't just make `/cloud/services/lightdb-stream` redirect elsewhere.

It may somewhat of a fool's errand to keep these redirects up to date, but I do anticipate some more churn coming around the docs with respect to application services, so I think I'd advocate for keeping these links for now until our documentation stabilizes.

sam-golioth: Sounds good to me!

sam-golioth (approved):

PR #505, opened 2024-05-29
fw_update: use backoff-delay for fw update observation retry

szczys:
When the mailbox is full, the firmware update observation will try to re-register. The increment for the retry count has been moved outside of the conditional tests so that the function will return when retries are exhausted.

resolves https://github.com/golioth/firmware-issue-tracker/issues/573


src/Kconfig
```diff
@@ -37,6 +37,16 @@ config GOLIOTH_OTA_MAX_NUM_COMPONENTS
     help
         Maximum number of components in an OTA manifest.
 
+config GOLIOTH_FW_OBSERVATION_RETRY_MAX_DELAY_S
+    int "Golioth maximum delay before retrying firmware update observation"
+    default 3600 # 1 hour
```
szczys, line 42:
```suggestion
    default 300 # 5 minutes
```

I tested this at 1 hour and it certainly works. But I think the default might be better at a lower value to that it's more likely a user will see the log message when troubleshooting issue. It is fully configurable when a longer value is needed in a project.

src/Kconfig
```diff
@@ -37,6 +37,16 @@ config GOLIOTH_OTA_MAX_NUM_COMPONENTS
     help
         Maximum number of components in an OTA manifest.
 
+config GOLIOTH_FW_OBSERVATION_RETRY_MAX_DELAY_S
```
sam-golioth, line 40:
```suggestion
config GOLIOTH_OTA_OBSERVATION_RETRY_MAX_DELAY_S
```
I think `FW` can be ambiguous, as it really just means "firmware".

src/Kconfig
```diff
@@ -37,6 +37,16 @@ config GOLIOTH_OTA_MAX_NUM_COMPONENTS
...
+    int "Golioth maximum delay before retrying firmware update observation"
```
sam-golioth, line 41:
```suggestion
    int "Golioth maximum delay before retrying OTA manifest observation"
```

src/Kconfig
```diff
@@ -37,6 +37,16 @@ config GOLIOTH_OTA_MAX_NUM_COMPONENTS
...
+    default 3600 # 1 hour
+    help
+        Maximum number of seconds to delay between firmware observation retries.
```
sam-golioth, line 44:
```suggestion
        Maximum number of seconds to delay between attempts to establish an OTA manifest observation.
```

src/Kconfig
sam-golioth: I'm fine either way. This is another config option that we'll probably want to provide some guidance around best practices eventually.

sam-golioth (approved):

PR #506, opened 2024-05-29
hil: ota: add test for observation after restart

szczys:
Test that the firmware update observation is reestablished after a client restart.

resolves https://github.com/golioth/firmware-issue-tracker/issues/547


tests/hil/tests/ota/test.c
```diff
@@ -168,6 +172,24 @@ static void test_block_ops(void)
...
+        golioth_client_stop(client);
+        golioth_sys_msleep(3000);
+        GLTH_LOGI(TAG, "Restarting Client");
+        golioth_sys_sem_give(restart_pending_sem);
+        golioth_client_start(client);
+    }
+    else if (golioth_sys_sem_take(restart_pending_sem, 0))
+    {
+        GLTH_LOGI(TAG, "Restart successful");
+        golioth_sys_sem_give(restart_test_begin_sem);
```
szczys, line 189:
```suggestion
```
We don't need to reset any semaphores for this test. We want each to run once and only once, to prevent restarting the test when the artifact is rolled back in the teardown phase of the test.

tests/hil/tests/ota/test.c
```diff
@@ -168,6 +172,24 @@ static void test_block_ops(void)
     }
 }
 
+static void test_restart(void)
+{
+
+    if (golioth_sys_sem_take(restart_test_begin_sem, 0))
```
sam-golioth, line 178: I don't think we need to use semaphores inside of this function. We're not performing mutual exclusion between multiple threads, and we're not signaling from one thread to another.

szczys: Sounds good, I've converted this to use a bool variable.

sam-golioth: This feels a lot more straightforward!

sam-golioth (approved):

szczys:
This PR is quite stale. I'm closing this for now, we can open a new PR if we end up picking this back up again.

PR #507, opened 2024-05-29
hil: settings: add test for observation after restart

szczys:
Test that the settings observation is reestablished after a client restart.

resolves https://github.com/golioth/firmware-issue-tracker/issues/546


tests/hil/tests/settings/test.c
```diff
@@ -2,11 +2,14 @@
...
 #include <golioth/golioth_sys.h>
 #include <golioth/settings.h>
+#include <string.h>
 
 LOG_TAG_DEFINE(test_settings);
 
 static golioth_sys_sem_t connected_sem;
 static golioth_sys_sem_t cancel_all_sem;
+static golioth_sys_sem_t restart_sem;
+static golioth_sys_sem_t restart_test_flag_sem;
```
sam-golioth, line 12: This is pretty confusing - I think I was able to figure it out eventually, but it's immediately not clear why we're using 2 semaphores, and their names aren't descriptive enough to distinguish them from each other. This is so that the test doesn't get re-triggered when observations are re-established, correct? Can we handle this the same way we do for the cancel observations test? By using a boolean and setting it back to false before the client reconnects. We should also confirm that the device is disconnected by attempting to change a setting, again similar to the cancel_all test.

sam-golioth (changes requested):

tests/hil/tests/settings/test.c
szczys: The problem I ran into in that approach is that the device can't be updated while the client is stopped. So the alternative to what I've done here is to stop the client and wait for an arbitrary amount of time, then start it again. Do you prefer that?

tests/hil/tests/settings/test.c
```diff
@@ -133,6 +151,26 @@ void hil_test_entry(const struct golioth_client_config *config)
...
+
+            golioth_client_stop(client);
+
+            golioth_sys_msleep(15 * 1000);
+
+            golioth_client_start(client);
+
+            golioth_sys_msleep(5 * 1000);
+
+            GLTH_LOGI(TAG, "Client restarted");
```
szczys, line 168:
```suggestion
            golioth_sys_msleep(1000);

            golioth_client_stop(client);
            golioth_sys_msleep(15 * 1000);

            golioth_client_start(client);
            golioth_sys_msleep(5 * 1000);
            GLTH_LOGI(TAG, "Client restarted");
```

Remove some of the blank lines.

tests/hil/tests/settings/test_settings.py
```diff
@@ -207,3 +208,24 @@ async def test_cancel_all(board, device):
...
+
+async def test_restart(board, device):
+    await device.settings.set('TEST_RESTART', True)
+    assert None != board.wait_for_regex_in_line('Received test_restart: true', timeout_s=10)
+    assert None != board.wait_for_regex_in_line('Ending session', timeout_s=10)
+
+    # Check that we no longer receive this settings change
+    await device.settings.set('TEST_INT', 2320)
+    with pytest.raises(RuntimeError) as e:
+        assert None != board.wait_for_regex_in_line('Received test_int', timeout_s=10)
```
sam-golioth, line 221: We can probably drop this timeout to 5 and reduce the 15 second timeout in the firmware to 10, if we're concerned about the length of the test. But I'm also fine with the current values.

sam-golioth (approved):

tests/hil/tests/settings/test_settings.py
```diff
@@ -207,3 +208,24 @@ async def test_cancel_all(board, device):
...
+    assert str(e.value) == "Timeout"
+
+    # Wait for client to restart
+    assert None != board.wait_for_regex_in_line('Client restarted', timeout_s=60)
+
+    # Wait for rush of initial settings log messages to pass
+    time.sleep(2)
```
mniestroj, line 228: Nit: I'd prefer `await trio.sleep(5)` instead. This won't change much in the execution, but it makes more sense to use async equivalent in async context, as well as for consistency.

mniestroj (approved):

tests/hil/tests/settings/test_settings.py
szczys: Yeah, thanks for the callout. I saw that PR you made with this and agree it's a better approach. I'll update the outstanding PRs before merging.

tests/hil/tests/settings/test_settings.py
szczys: I'm also going to change this to test for the number that was set:

```suggestion
        assert None != board.wait_for_regex_in_line('Received test_int: 2320', timeout_s=10)
```

PR #510, opened 2024-06-03
ci: don't fail fast

sam-golioth:
Runs all HIL tests, even if one fails, instead of exiting after the first failure.

A test run in CI where I forced the OTA test to fail is here: https://github.com/golioth/golioth-firmware-sdk/actions/runs/9354727595

Resolves golioth/firmware-issue-tracker#577


szczys (approved):
Clever!

mniestroj (approved):

PR #512, opened 2024-06-05
Update certificate_provisioning test so it can run on both dev and prod

szczys:
Get the project id from the project fixture in pytest instead of using a hardcoded value. This allows the test to run on prod or dev.

This was tested against dev and found to be working: https://github.com/golioth/golioth-firmware-sdk/actions/runs/9387861576

resolves https://github.com/golioth/firmware-issue-tracker/issues/561


sam-golioth (approved):
Nice change!

mniestroj (approved):

PR #515, opened 2024-06-06
coap_blockwise: use non-blocking APIs for CoAP operations

sam-golioth:
The blocking APIs in the CoAP client create and destroy a semaphore each time they are called. For a complete blockwise transfer, we could call these APIs thousands of times. To avoid creating and destroying a semaphore each time, we instead manage the blocking in the blockwise transfers module, which now uses the non-blocking APIs. This allows us to create and destroy the semaphore just once for the complete blockwise transfer, rather than for each block.

Also, rename the `status` member of the context struct to `response_status` to better reflect how it is used.

Resolves golioth/firmware-issue-tracker#581

### DIFF
--- a/src/coap_blockwise.c
+++ b/src/coap_blockwise.c
@@ -13,8 +13,9 @@
 struct blockwise_transfer
 {
     bool is_last;
-    enum golioth_status status;
+    enum golioth_status response_status;
     enum golioth_content_type content_type;
+    golioth_sys_sem_t sem;
     uint32_t offset;
     size_t block_size;
     const char *path_prefix;
@@ -58,7 +59,9 @@ static void on_block_sent(struct golioth_client *client,
 {
     assert(arg);
     struct blockwise_transfer *ctx = (struct blockwise_transfer *) arg;
-    ctx->status = response->status;
+    ctx->response_status = response->status;
+
+    golioth_sys_sem_give(ctx->sem);
 }
 
 // Function to call the application's read block callback after a successful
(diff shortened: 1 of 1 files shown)


src/coap_blockwise.c
```diff
@@ -225,23 +251,32 @@ static void on_block_rcvd(struct golioth_client *client,
...
+                                                            ctx->content_type,
+                                                            ctx->offset,
+                                                            ctx->block_size,
+                                                            on_block_rcvd,
+                                                            ctx,
+                                                            false,
+                                                            GOLIOTH_SYS_WAIT_FOREVER);
+    if (GOLIOTH_OK == err)
+    {
+        golioth_sys_sem_take(ctx->sem, GOLIOTH_SYS_WAIT_FOREVER);
```
szczys, line 275: I'm having a bit of trouble following how this change works. Will block downloads always block (prevent client from doing anything else until another chunk is received) or will there be an async option as well?

szczys (approved):

src/coap_blockwise.c
szczys: Ah, I see this is going to be run on the [fw_update thread](https://github.com/golioth/golioth-firmware-sdk/blob/2e09cda7163d15212706ccfe24c1571ae7e5e396/src/fw_update.c#L170) so the blocking will happen there an not on the client thread.

src/coap_blockwise.c
```diff
@@ -81,18 +84,25 @@ static enum golioth_status upload_single_block(struct golioth_client *client,
...
+                                                            ctx->is_last,
+                                                            GOLIOTH_CONTENT_TYPE_JSON,
+                                                            ctx->offset,
+                                                            ctx->block_buffer,
+                                                            ctx->block_size,
+                                                            on_block_sent,
+                                                            ctx,
+                                                            false,
+                                                            GOLIOTH_SYS_WAIT_FOREVER);
+    if (GOLIOTH_OK == err)
```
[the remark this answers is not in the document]
sam-golioth: It's a defensive pattern. It you make a typo and write `=` instead of `==`, placing the constant first will cause a compiler error, while placing the variable first will compile correctly but probably isn't what you intended.

src/coap_blockwise.c
sam-golioth: Yep, exactly right. Generally it will happen on the user thread, whatever that happens to be. I would like to get an asynchronous version later, where all operations happen on the client thread, but that's for the future.

PR #516, opened 2024-06-10
zephyr: coap: periodically report resend count

szczys:
Sum the number of coap request resends over a period of time and report as an INF log message. Change the existing resend message to DBG level so it is still available when that log level is selected via Kconfig.

I'm proposing this approach as an alternative to #500. We had originally discussed changing the log level of `zephyr_coap_req.c` in only the golioth log backend so the messages are reported in the terminal but not sent to the cloud. This has two disadvantages:
- Setting a runtime level for that backend felt a little bit convoluted as it needs to be done after the backend is initialized and we don't initialize until after we connect to Golioth
- When only a few resend messsages occur, they're pretty handy to have (even on remote logs).

This PR is a rather small resource change (adds 6 bytes to the Golioth client) and low overhead (we already have the uptime value so it's just a some subtraction/multiplication on each conditional).

resolves https://github.com/golioth/firmware-issue-tracker/issues/576
closes #500

### DIFF
--- a/src/zephyr_coap_req.c
+++ b/src/zephyr_coap_req.c
@@ -17,6 +17,8 @@ LOG_MODULE_DECLARE(golioth_coap_client);
 
 static const int64_t COAP_OBSERVE_TS_DIFF_NEWER = 128 * (int64_t) MSEC_PER_SEC;
 
+#define RESEND_REPORT_TIMEFRAME_S 10
+
 #define COAP_RESPONSE_CODE_CLASS(code) ((code) >> 5)
 
 void golioth_coap_reqs_init(struct golioth_client *client)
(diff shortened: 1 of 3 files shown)


src/zephyr_coap_req.c
```diff
@@ -699,10 +701,12 @@ static int64_t golioth_coap_req_poll_prepare(struct golioth_coap_req *req, uint3
     {
         if (resend)
         {
-            LOG_WRN("Resending request %p (reply %p) (retries %d)",
+            LOG_DBG("Resending request %p (reply %p) (retries %d)",
```
sam-golioth, line 704: This should still be a warning.

sam-golioth (approved):

src/zephyr_coap_req.c
szczys: I think you meant to leave this comment on line 722, is that right?

```
LOG_INF("%u resends in last %d seconds",
```

PR #520, opened 2024-06-11
coap client: handle allocation errors with sync calls

szczys:
- When creating an event group, check for memory allocation and semaphore creation errors. Free memory as needed and return NULL
- During sync calls, check event group and sempahore for NULL. Free memory as needed and return GOLIOTH_ERROR_MEM_ALLOC

resolves https://github.com/golioth/firmware-issue-tracker/issues/582

### DIFF
--- a/src/coap_client.c
+++ b/src/coap_client.c
@@ -82,7 +82,6 @@ enum golioth_status golioth_coap_client_empty(struct golioth_client *client,
         {
             golioth_event_group_destroy(request_msg.request_complete_event);
             golioth_sys_sem_destroy(request_msg.request_complete_ack_sem);
-            request_msg.status = NULL;
         }
         return GOLIOTH_ERR_QUEUE_FULL;
     }
@@ -232,7 +231,6 @@ static enum golioth_status golioth_coap_client_set_internal(struct golioth_clien
         {
             golioth_event_group_destroy(request_msg.request_complete_event);
             golioth_sys_sem_destroy(request_msg.request_complete_ack_sem);
-            request_msg.status = NULL;
         }
         return GOLIOTH_ERR_QUEUE_FULL;
     }
@@ -398,7 +396,6 @@ enum golioth_status golioth_coap_client_delete(struct golioth_client *client,
         {
             golioth_event_group_destroy(request_msg.request_complete_event);
             golioth_sys_sem_destroy(request_msg.request_complete_ack_sem);
-            request_msg.status = NULL;
         }
         return GOLIOTH_ERR_QUEUE_FULL;
     }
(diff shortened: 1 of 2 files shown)


src/coap_client.c
```diff
@@ -421,10 +448,19 @@ static enum golioth_status golioth_coap_client_get_internal(struct golioth_clien
...
+        // Created here, deleted by coap thread (or here if fail to create or enqueue)
         request_msg.request_complete_event = golioth_event_group_create();
         request_msg.request_complete_ack_sem = golioth_sys_sem_create(1, 0);
         request_msg.status = &status;
+
+        if (!request_msg.request_complete_event || !request_msg.request_complete_ack_sem)
+        {
+            GLTH_LOGW(TAG, "Failed to create request_complete members");
+            golioth_event_group_destroy(request_msg.request_complete_event);
+            golioth_sys_sem_destroy(request_msg.request_complete_ack_sem);
```
sam-golioth, line 460: We can't assume that `golioth_sys_sem_destroy()` is tolerant to being called with a semaphore that did not create successfully. For example. FreeRTOS will assert if you attempt to destroy a NULL semaphore, and the ESP-IDF port does not check for NULL.

We should just check each one of these for failure after creating it. There's not an advantage in checking them together other than reducing the lines of code, but that's not an advantage in itself and in this case I believe having separate checks is more readable and maintainable.

sam-golioth (changes requested):

src/coap_client.c
```diff
@@ -162,9 +177,24 @@ static enum golioth_status golioth_coap_client_set_internal(struct golioth_clien
 
     if (is_synchronous)
     {
-        // Created here, deleted by coap thread (or here if fail to enqueue
+        // Created here, deleted by coap thread (or here if fail to create or enqueue)
         request_msg.request_complete_event = golioth_event_group_create();
+        if (!request_msg.request_complete_event)
+        {
+            GLTH_LOGW(TAG, "Failed to create event group");
+            request_msg.status = NULL;
```
sam-golioth, line 185: Same comment as above, this isn't necessary (throughout this, the same is true).

src/coap_client.c
```diff
@@ -162,9 +177,24 @@ static enum golioth_status golioth_coap_client_set_internal(struct golioth_clien
...
+            return GOLIOTH_ERR_MEM_ALLOC;
```
sam-golioth, line 186: We need to free `request_payload` allocated on line 160 above before returning.

sam-golioth (changes requested):

src/coap_client.c
szczys: Great, I've made a subsequent commit to remove similar code not previously touched by this PR.

src/coap_client.c
szczys: Good catch, thank you!

I made a subsequent commit to change how this is tested a few lines before this.

sam-golioth (approved):
Looks good!

PR #529, opened 2024-06-24
zephyr: coap: fix keepalive behavior

szczys:
- ensure keepalive timer is reset after callback
- reset keepalive timer whenever a response is received
- remove ping timer logic and Kconfig symbol

The ping timer was duplicating the keepalive timer functionality. The keepalive is used by the libcoap platforms so this change brings better parity.

resolves https://github.com/golioth/firmware-issue-tracker/issues/592


mniestroj (approved):

szczys:
I verified that the keepalive is being sent through the mbox so we should be good in terms of thread-safeness. Moving this out of draft.

PR #533, opened 2024-07-03
Add extern “C” block to header files

MarkoPura:
By adding an extern "C" block to header files, we avoid linking problems when building C++ applications that come from name mangling since C++ supports function overloading.

### DIFF
--- a/include/golioth/client.h
+++ b/include/golioth/client.h
@@ -3,6 +3,12 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
 #pragma once
 
 #include <stdbool.h>
@@ -286,3 +292,7 @@ void golioth_client_set_packet_loss_percent(uint8_t percent);
 golioth_sys_thread_t golioth_client_get_thread(struct golioth_client *client);
 
 /// @}
+
+#ifdef __cplusplus
+}
+#endif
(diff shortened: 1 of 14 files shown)


mniestroj (commented):
I get why `include/golioth/*.h` files need that. But how about `src/*.h`? Those are private and should not be built with C++ compiler.

MarkoPura:
I agree @mniestroj and was debating with myself whether to add it to `src/*.h` files or not. I've left it there in the end since it "doesn't hurt", but I can always remove it.

src/coap_blockwise.h
```diff
@@ -3,6 +3,12 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
+
+#ifdef __cplusplus
```
sam-golioth, line 7: Is this also required in the private headers? Or only in the public headers? These headers in `src/` aren't available to user applications, so they will only be included by C files, not C++.

MarkoPura: Will remove from the private headers in `src/`

mniestroj:
> I agree @mniestroj and was debating with myself whether to add it to `src/*.h` files or not. I've left it there in the end since it "doesn't hurt", but I can always remove it.

IMO it does hurt :) Any additional code that is not used is a dead code. Dead code not tested and potentially not working. Even if that is only so small, `#ifdef __cplusplus` will never be true.

So I opt for removing it from private headers.

MarkoPura:
@sam-golioth, @mniestroj removed from the `src/` folder!

sam-golioth (approved):
Thanks @MarkoPura!

mniestroj (approved):

PR #536, opened 2024-07-10
Add more documentation to read coap block callback

szczys:
- stream: update doxygen for read block callback

### DIFF
--- a/include/golioth/stream.h
+++ b/include/golioth/stream.h
@@ -72,15 +72,24 @@ enum golioth_status golioth_stream_set_sync(struct golioth_client *client,
 
 /// Read block callback
 ///
-/// This callback will be called by the golioth client each time it needs to
+/// This callback will be called by the Golioth client each time it needs to
 /// fill a block.
 ///
+/// When the callback runs, the buffer should be filled with `block_size` number
+/// of bytes. When the last block contains fewer bytes, set `block_size` to the
+/// actual number of bytes written. Set `is_last` to indicate all data has been
+/// written to buffer.
+///
 /// @param block_idx The index of the block to fill
-/// @param block_buffer The buffer that this callabck should fill
+/// @param block_buffer The buffer that this callback should fill
 /// @param block_size (in/out) Contains the maximum size of block_buffer, and
-///        should be set to the size of data actually placed in block_buffer
+///        should be set to the size of data actually placed in block_buffer.
+///        Value must be > 0 when this function returns GOLIOTH_OK.
 /// @param is_last (out) Set this to true if this is the last block to transfer
 /// @param arg A user provided argument
+///
+/// @retval GOLIOTH_OK Callback ran successfully
+/// @retval GOLIOTH_ERR_* Indicate error type encountered by callback
 typedef enum golioth_status (*stream_read_block_cb)(uint32_t block_idx,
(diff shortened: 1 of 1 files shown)


src/coap_blockwise.c
```diff
@@ -142,9 +142,9 @@ static enum golioth_status process_blockwise_uploads(struct golioth_client *clie
 {
     enum golioth_status status = GOLIOTH_ERR_FAIL;
     int ret = call_read_block_callback(ctx);
```
sam-golioth, line 144: This function does actually return a negative value, so the assert you hit must be caused by something else.

include/golioth/stream.h
```diff
@@ -72,13 +72,18 @@ enum golioth_status golioth_stream_set_sync(struct golioth_client *client,
...
 /// @param block_idx The index of the block to fill
-/// @param block_buffer The buffer that this callabck should fill
+/// @param block_buffer The buffer that this callback should fill
 /// @param block_size (in/out) Contains the maximum size of block_buffer, and
-///        should be set to the size of data actually placed in block_buffer
+///        should be set to the size of data actually placed in block_buffer.
+///        Value must always be > 0.
 /// @param is_last (out) Set this to true if this is the last block to transfer
 /// @param arg A user provided argument
 typedef enum golioth_status (*stream_read_block_cb)(uint32_t block_idx,
```
sam-golioth, line 89: Let's add information about the return value.

include/golioth/stream.h
```diff
@@ -72,13 +72,18 @@ enum golioth_status golioth_stream_set_sync(struct golioth_client *client,
 
 /// Read block callback
 ///
-/// This callback will be called by the golioth client each time it needs to
+/// This callback will be called by the Golioth client each time it needs to
 /// fill a block.
 ///
+/// When the callback runs, the buffer should be filled with `block_size` number of bytes. When the
```
sam-golioth, line 78: Let's wrap at 80 to match the rest of this file

include/golioth/stream.h
```diff
@@ -72,13 +72,18 @@ enum golioth_status golioth_stream_set_sync(struct golioth_client *client,
...
+/// `last block` contains fewer bytes, set `block_size` to the actual number of bytes written. Set
+/// `is_last` to indicate all data has been written to buffer.
+///
 /// @param block_idx The index of the block to fill
-/// @param block_buffer The buffer that this callabck should fill
+/// @param block_buffer The buffer that this callback should fill
 /// @param block_size (in/out) Contains the maximum size of block_buffer, and
-///        should be set to the size of data actually placed in block_buffer
+///        should be set to the size of data actually placed in block_buffer.
+///        Value must always be > 0.
```
sam-golioth, line 86:
```suggestion
///        Value must be > 0 when this function returns GOLIOTH_OK,
```

sam-golioth (changes requested):

src/coap_blockwise.c
szczys: Indeed, I tested this as working but if I return `GOLIOTH_OK` it asserts:

```
    if (block_idx == 10)
    {
        *block_size = 0;
        *is_last = true;
        return GOLIOTH_OK;
    }
```

sam-golioth (approved):

PR #486, opened 2024-05-08
ESP log backend

rashmivaidya72:

PR #547, opened 2024-07-30
Stop client before destroying

szczys:
- Fix memory free for post_block requests in both libcoap and Zephyr coap
- Stop client before destroying it in both libcoap and Zephyr coap
- libcoap: call golioth_sys_client_disconnected() when ending a session
- libcoap: disable automatic cloud logging when the client is disconnected

resolves https://github.com/golioth/firmware-issue-tracker/issues/584

### DIFF
--- a/port/freertos/golioth_sys_freertos.c
+++ b/port/freertos/golioth_sys_freertos.c
@@ -14,11 +14,13 @@
  * Time
  *------------------------------------------------*/
 
-uint64_t golioth_sys_now_ms() {
+uint64_t golioth_sys_now_ms()
+{
     return xTaskGetTickCount() * portTICK_PERIOD_MS;
 }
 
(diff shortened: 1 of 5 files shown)


port/freertos/golioth_sys_freertos.c
```diff
@@ -14,38 +14,45 @@
  * Time
  *------------------------------------------------*/
 
-uint64_t golioth_sys_now_ms() {
+uint64_t golioth_sys_now_ms()
```
sam-golioth, line 17: These are moving from the FreeRTOS port to the system ports (ESP-IDF, ModusToolBox, etc.) in https://github.com/golioth/golioth-firmware-sdk/pull/486, but I think it's fine to keep it here for now.

sam-golioth (approved):
Great work @szczys!

PR #548, opened 2024-07-31
Remove ota compression and patch

szczys:
The compression and patch functionality of the Golioth Firmware SDK are not yet supported on the server side. These features may be added back at some point in the future when end-to-end support becomes available.

resolves https://github.com/golioth/firmware-issue-tracker/issues/615


sam-golioth (approved):
🎉 Just a reminder to hold the merge until after v0.15.0. Thanks @szczys!

mniestroj (approved):

PR #554, opened 2024-08-06
Enable p384 curve

sam-golioth:
This is the minimal set of changes to enable support for the P-384 curve in the Zephyr based ports. We may wish to enable further curves and algorithms at a later date, but for now this immediately restores certificate functionality.

TODO:
- [x] Test with offloaded sockets on nRF9160dk

Resolves golioth/firmware-issue-tracker#622

### DIFF
--- a/examples/zephyr/certificate_provisioning/boards/nrf9160dk_nrf9160_ns.conf
+++ b/examples/zephyr/certificate_provisioning/boards/nrf9160dk_nrf9160_ns.conf
@@ -28,5 +28,11 @@ CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=2048
 CONFIG_MBEDTLS_KEY_EXCHANGE_DHE_PSK_ENABLED=n
 CONFIG_MBEDTLS_KEY_EXCHANGE_DHE_RSA_ENABLED=n
 
+# MbedTLS configuration to support p-384 curve. These options
+# enable using the MbedTLS built-in support for operations not
+# supported by the default nRF Oberon crypto backend
+CONFIG_NORDIC_SECURITY_BACKEND=n
+CONFIG_MBEDTLS_LEGACY_CRYPTO_C=y
+
 # Generate MCUboot compatible images
 CONFIG_BOOTLOADER_MCUBOOT=y
--- a/port/zephyr/Kconfig
+++ b/port/zephyr/Kconfig
@@ -59,8 +59,10 @@ config GOLIOTH_AUTH_PSK_MBEDTLS_DEPS
 config GOLIOTH_AUTH_CERT_MBEDTLS_DEPS
 	bool "mbedTLS dependencies for certificate based auth"
 	depends on MBEDTLS_BUILTIN || NRF_SECURITY
-	# At least one ECP needs to be selected. This was chosen arbitrarily.
+	# Enable P-256 and P-384 for broad compatibility. Disabling these curves
+	# may cause incompatibility with Golioth servers, now or in the future.
 	imply MBEDTLS_ECP_DP_SECP256R1_ENABLED
+	imply MBEDTLS_ECP_DP_SECP384R1_ENABLED
 	# Select at least one server supported cipher
 	imply MBEDTLS_CIPHER_GCM_ENABLED if (MBEDTLS_BUILTIN && !MBEDTLS_CIPHER_CCM_ENABLED)
 	select MBEDTLS_GCM_C if (NRF_SECURITY && !MBEDTLS_CCM_C)
--- a/examples/zephyr/fw_update/boards/nrf9160dk_nrf9160_ns.conf
+++ b/examples/zephyr/fw_update/boards/nrf9160dk_nrf9160_ns.conf
@@ -28,5 +28,11 @@ CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=2048
 CONFIG_MBEDTLS_KEY_EXCHANGE_DHE_PSK_ENABLED=n
 CONFIG_MBEDTLS_KEY_EXCHANGE_DHE_RSA_ENABLED=n
 
+# MbedTLS configuration to support p-384 curve. These options
+# enable using the MbedTLS built-in support for operations not
+# supported by the default nRF Oberon crypto backend
+CONFIG_NORDIC_SECURITY_BACKEND=n
+CONFIG_MBEDTLS_LEGACY_CRYPTO_C=y
+
 # Generate MCUboot compatible images
 CONFIG_BOOTLOADER_MCUBOOT=y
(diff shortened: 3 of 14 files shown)


examples/zephyr/certificate_provisioning/boards/nrf9160dk_nrf9160_ns.conf
```diff
@@ -28,5 +28,11 @@ CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=2048
 CONFIG_MBEDTLS_KEY_EXCHANGE_DHE_PSK_ENABLED=n
 CONFIG_MBEDTLS_KEY_EXCHANGE_DHE_RSA_ENABLED=n
 
+# MbedTLS configuration to support p-384 curve. These options
+# enable using the MbedTLS built-in support for operations not
+# supported by the default nRF Oberon crypto backend
+CONFIG_NORDIC_SECURITY_BACKEND=n
+CONFIG_MBEDTLS_LEGACY_CRYPTO_C=y
```
sam-golioth, line 35: @mniestroj I'd like your thoughts on if this is the right place for these configuration options. I put them here to match the above options, but I'm wondering if it would make sense to set these defaults in a more central location so that customer applications pick these up automatically, since they are essentially required for all applications using certificate authentication. The nordic-specific nature of these options gave me pause.

mniestroj: The only alternative solution I see is utilizing [snippets](https://docs.zephyrproject.org/latest/build/snippets/index.html), so that our customers can **optionally** use the configurations known to work, without doing everything at the application level.

For now (until we give snippets a try), I think it makes sense to leave this configuration here.

mniestroj: My gut feeling is that this configuration will not be supported for a long time and we might need to search for new working Kconfig options every time we bump NCS version. The option set in this file is just getting more and more specific, which is more and more likely to be broken (by dropping support for any subset of it) in the future.

mniestroj: What are the implications of above options for the default PSK configuration in terms of resource usage?

sam-golioth: > so that our customers can optionally use the configurations known to work

One of the things I'm worried about here is that these aren't really optional. There could be other ways of supporting the same curve (like through PSA instead of MbedTLS), so I don't want to make them y-selects, but if customers don't have support for this curve, their devices won't work. That's probably enough of a barrier for these particular options (users will immediately see that something is broken and try to fix it). But the issues we're having now are making me think about how we can make sure that users have support for curves and algorithms they might need in the future (i.e. how could we have enforced support for the P-384 curve in the past even what it wasn't used, to make sure devices would have been compatible when we switched over).

> My gut feeling is that this configuration will not be supported for a long time and we might need to search for new working Kconfig options every time we bump NCS version.

To some extent, that's just something we're going to have deal with. Nordic can change things at any time and we're always going to need to make changes to match. It seems like they're moving towards using the PSA API as the crypto backend though, and I'm hopeful that will be more stable. I think we could even make that switch now, but it's a bigger change so I'd rather make the minimal changes here and look at moving to PSA in a later release.

> What are the implications of above options for the default PSK configuration in terms of resource usage?

Looks like about 600 extra bytes of flash in the hello sample:
On main:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:      293124 B       448 KB     63.90%
             RAM:       66364 B     211608 B     31.36%
        IDT_LIST:          0 GB         2 KB      0.00%
```

On this branch:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:      292524 B       448 KB     63.77%
             RAM:       66364 B     211608 B     31.36%
        IDT_LIST:          0 GB         2 KB      0.00%
```

mniestroj (approved):

port/zephyr/Kconfig
```diff
@@ -59,8 +59,9 @@ config GOLIOTH_AUTH_PSK_MBEDTLS_DEPS
 config GOLIOTH_AUTH_CERT_MBEDTLS_DEPS
 	bool "mbedTLS dependencies for certificate based auth"
 	depends on MBEDTLS_BUILTIN || NRF_SECURITY
-	# At least one ECP needs to be selected. This was chosen arbitrarily.
+	# Leaf certs use P-256, Intermediate and Root certs use P-384
```
hasheddan, line 62: nitpick: @sam-golioth can we say something like `Enable support for both P-256 and P-384 curves for broad compatibility` or similar here rather than tying this description to the current trust chain on the server?

examples/zephyr/certificate_provisioning/boards/nrf9160dk_nrf9160_ns.conf
hasheddan: One other thing to note here (which I don't think should impact current decision making) is that in addition to impact on footprint, I believe that using the mbedtls built-in rather than nordic backends means that we miss out on any hardware-specific acceleration, which could impact runtime performance. Would be good to include analysis of this impact in future optimization documentation 👍🏻

sam-golioth: I'm not actually sure that this does disable hardware acceleration (would need to investigate). The 9160 datasheet advertises support for the p-384 curve in acceleration, so it's odd that we would need to disable acceleration to enable the curve. The crypto options in NCS are very confusing however, so it's definitely possible. As far as I can tell, there are at least 3 crypto backends available, and at least 2 of those are MbedTLS.

I think where we want to get to eventually is having explicit documentation on which curves and algorithms need to be supported, then users can have the freedom (and confidence) to select the configuration and hardware that supports those requirements and matches their other needs.

hasheddan: > I'm not actually sure that this does disable hardware acceleration (would need to investigate). The 9160 datasheet advertises support for the p-384 curve in acceleration, so it's odd that we would need to disable acceleration to enable the curve.

I don't think it disables hardware acceleration, but rather that it may not take full advantage of it. I think this is the primary purpose of the [utilizing the oberon backend](https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/libraries/security/nrf_security/doc/drivers.html#nrfoberon_driver):

> The [nrf_oberon crypto library](https://docs.nordicsemi.com/bundle/ncs-latest/page/nrfxlib/crypto/doc/nrf_oberon.html#nrf-oberon-readme) is distributed as a closed-source binary that provides select cryptographic algorithms optimized for use in nRF devices. This provides faster execution than the original Mbed TLS implementation.

That being said, because oberon doesn't support secp384 it is kind of a moot point anyway 👍🏻

sam-golioth: I think the oberon library is assembly-optimized but not hardware accelerated. MbedTLS should be able to take advantage of hardware acceleration. But it's still very confusing to me lol.

sam-golioth:
I've confirmed that this PR also works with offloaded DTLS sockets on the nRF9160DK without further changes

hasheddan (approved):

PR #555, opened 2024-08-07
Allow configurable block size for blockwise upload and download

szczys:
- Harden the configuration options for both upload and download blocksize by constraining Kconfig symbols to values allowed by the CoAP spec and asserting those allowed values at build time.
- Honor the blocksize Kconfig setting for blockwise download (from the device side). This was already the case for blockwise upload.

This PR does not implement the case where the server response indicates a smaller block size than the original request from the device indicated. That will be implemented in a future PR.  

resolve https://github.com/golioth/firmware-issue-tracker/issues/600


src/coap_blockwise.c
szczys, line 1: Line 239 has the wrong symbol:

```
- ctx->block_size = CONFIG_GOLIOTH_BLOCKWISE_UPLOAD_MAX_BLOCK_SIZEl
+ ctx->block_size = CONFIG_GOLIOTH_BLOCKWISE_DOWNLOAD_MAX_BLOCK_SIZE;

File: src/Kconfig.coap [hunk 1]
```diff
@@ -40,12 +40,42 @@ config GOLIOTH_MAX_NUM_OBSERVATIONS
...
+        bool "128"
+    config UP_BLOCKSIZE_64
+        bool "64"
+    config UP_BLOCKSIZE_32
+        bool "32"
+    config UP_BLOCKSIZE_16
+        bool "16"
+
     help
         Block size in Blockwise Uploads.
```
mniestroj, line 94: Shoudn't `help` be just after `prompt`, before any choices listed with `config`?

src/Kconfig.coap
```diff
@@ -40,12 +40,42 @@ config GOLIOTH_MAX_NUM_OBSERVATIONS
     help
         The maximum number of CoAP paths which can be simultaneously observed.
 
-config GOLIOTH_BLOCKWISE_UPLOAD_BLOCK_SIZE
-    int "Golioth blockwise uploads block size"
-    default 1024
+choice
```
mniestroj, line 43: Let's add a name like `choice UP_BLOCKSIZE`, so it is possible to override that value with new Kconfig default (e.g. application specific).

src/Kconfig.coap
```diff
@@ -40,22 +40,54 @@ config GOLIOTH_MAX_NUM_OBSERVATIONS
     help
         The maximum number of CoAP paths which can be simultaneously observed.
 
+choice
+    prompt "Golioth blockwise download: Max block size"
```
mniestroj, line 44: Let's add a name like `choice GOLIOTH_BLOCKSIZE_DN` to be able to override default value on application-level.

src/Kconfig.coap
```diff
@@ -40,22 +40,54 @@ config GOLIOTH_MAX_NUM_OBSERVATIONS
...
+    default 16 if GOLIOTH_BLOCKSIZE_DN_16
+    help
+        Buffer size used in blockwise downloads. The block download size negotiated with the server
+        will be no larger than the value of this setting.
+
 choice
     prompt "Golioth blockwise upload: Max block size"
 
-    config UP_BLOCKSIZE_1024
+    config GOLIOTH_BLOCKSIZE_UP_1024
```
mniestroj: Please rename that in the first commit instead.

src/Kconfig.coap
```diff
@@ -40,8 +40,56 @@ config GOLIOTH_MAX_NUM_OBSERVATIONS
...
+        bool "512"
+    config GOLIOTH_BLOCKSIZE_UP_256
+        bool "256"
+    config GOLIOTH_BLOCKSIZE_UP_128
+        bool "128"
+    config GOLIOTH_BLOCKSIZE_UP_64
+        bool "64"
+    config GOLIOTH_BLOCKSIZE_UP_32
+        bool "32"
+    config GOLIOTH_BLOCKSIZE_UP_16
```
sam-golioth, line 90: I think something went wonky with the rebase here, there's this list of `GOLIOTH_BLOCKSIZSE_UP_*` here but also still the list of `GOLIOTH_UP_BLOCKSIZE_*` below

sam-golioth (approved):

src/Kconfig.coap
szczys: Thanks for catching. I must have squashed into the wrong commit. All fixed now and I improved the help messages while I was in there.

PR #557, opened 2024-08-07
Use file settings backend in cert example

sam-golioth:
This changes the certificate provisioning sample to use the filesystem as the storage backend for persistent settings. This allows us to store wifi credentials alongside the client certificate and the private key in the filesystem, which in turn enables us to test WiFi devices in CI. Note that the certificate provisioning sample will continue to fail until #554 is merged, but a successful HIL run that includes those changes as well as the ones in this PR can be seen [here](https://github.com/golioth/golioth-firmware-sdk/actions/runs/10288446016/job/28474275154#step:8:49).

Resolves golioth/firmware-issue-tracker#391


mniestroj (approved):
Awesome!

PR #558, opened 2024-08-07
examples: esp-idf: enable SHA384 and SHA512 support

sam-golioth:
SHA-384 support is required for Certificate authentication. We enable it in all samples, because the choice of PSK vs Certificate auth is made at runtime, and this ensures that customers do not get confused when selecting a sample on which to base to their application.

This setting had been previously disabled to save space, but it is now required.

Resolves golioth/firmware-issue-tracker#626


hasheddan (approved):
Thanks @sam-golioth! I think enabling by default, then providing folks with tools to optimize accompanied by explicit statements of the consequences of doing so is the right move for the samples 👍🏻

PR #560, opened 2024-08-08
No reset in cert provisioning sample

sam-golioth:
This changes the certificate provisioning sample to repeatedly check for the presence of PKI credentials in the filesystem, instead of just once at boot as it previously did. This allows users (and more importantly, CI) to run through the example without needing to reset the board. Board that use a virtual COM port over a USB peripheral (like the RAK5010) will drop their serial connection if they reboot, so this change is necessary in order to test this sample in CI on those boards. Note that the certificate provisioning test will continue to fail until #554 is merged, but a passing run with the changes from that PR and the changes in this one can be found [here](https://github.com/golioth/golioth-firmware-sdk/actions/runs/10293915260/job/28491326370#step:8:49).

Resolves golioth/firmware-issue-tracker#627

### DIFF
--- a/examples/zephyr/certificate_provisioning/src/main.c
+++ b/examples/zephyr/certificate_provisioning/src/main.c
@@ -78,32 +78,32 @@ static int load_credential_from_fs(const char *path, uint8_t **buf_p, size_t *bu
         goto finish;
     }
 
-    /* NOTE: cred_buf is used directly by the TLS Credentials library, and so must remain
+    /* NOTE: *buf_p is used directly by the TLS Credentials library, and so must remain
      * allocated for the life of the program.
      */
 
-    void *cred_buf = malloc(dirent.size);
+    free(*buf_p);
+    *buf_p = malloc(dirent.size);
 
-    if (cred_buf == NULL)
+    if (*buf_p == NULL)
     {
         LOG_ERR("Could not allocate space to read credential");
         err = -ENOMEM;
         goto finish_with_file;
     }
 
-    err = fs_read(&file, cred_buf, dirent.size);
+    err = fs_read(&file, *buf_p, dirent.size);
 
     if (err < 0)
     {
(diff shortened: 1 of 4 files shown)


examples/zephyr/certificate_provisioning/src/main.c
```diff
@@ -140,13 +142,23 @@ int main(void)
...
+    {
+        int ret = load_credential_from_fs(CLIENT_CERTIFICATE_PATH, &tls_client_crt, &tls_client_crt_len);
+        if (ret > 0)
+        {
+            ret = load_credential_from_fs(PRIVATE_KEY_PATH, &tls_client_key, &tls_client_key_len);
+            if (ret > 0)
+            {
+                break;
+            }
+        }
```
mniestroj, line 155: What if the first cert will be successfully loaded and the second one will fail? We'll be stuck in this loop forever, right?

sam-golioth: Yeah that'e the intention. We should hang in this loop until the use has loaded both certificates. But this comment reminded me I need to update the README, so I'll do that next!

mniestroj: What about that case:
- `CLIENT_CERTIFICATE_PATH` is loaded successfully
- `PRIVATE_KEY_PATH` fails to load

in the next loop `CLIENT_CERTIFICATE_PATH` load will be attempted, but will most likely fail due to already being loaded. Or will it behave differently?

sam-golioth: That's a good callout! It will load fine, but we'll leak memory. I'll fix it 👍

sam-golioth: Fixed

mniestroj (approved):

PR #561, opened 2024-08-08
libcoap: use PEM instead of DER; add x2 CA cert

sam-golioth:
This commit adds the ISRG Root X2 cert, in addition to the existing ISRG Root X1 cert, for libcoap ports. To accomplish this, we need to switch back to using PEM formatted certs.

Resolves golioth/firmware-issue-tracker#625


scripts/certificates/generate_device_certificate.sh
```diff
@@ -2,8 +2,8 @@
 
 set -eu
 
-if [[ $# -ne 2 ]]; then
-    echo "usage: $0 <project_id> <certificate_id>" >&2
+if [[ $# -lt 2 ]]; then
+    echo "usage: $0 <project_id> <certificate_id> [pem]" >&2
```
szczys, line 6: I'm unfamiliar with the square-brackets syntax here (compared to the angle braces). Is this an argument in itself or is the user meant to replace that?

szczys (approved):

mniestroj (approved):

scripts/certificates/generate_device_certificate.sh
sam-golioth: Square brackets mean it's optional. The user should input "pem" exactly. If it's meant to be replaced and optional, then you would use both: `[<format>]`

PR #564, opened 2024-08-21
Add mutex API

szczys:
- Add mutex API
- Implement mutex in esp-idf, linux, zephyr

edit: I retargeted this onto #565 so that auto-formating is contained to its own PR and not show in this or subsequent work.

resolves https://github.com/golioth/firmware-issue-tracker/issues/631

### DIFF
--- a/port/linux/golioth_sys_linux.c
+++ b/port/linux/golioth_sys_linux.c
@@ -48,6 +48,61 @@ uint64_t golioth_sys_now_ms(void)
 }
 
 /*--------------------------------------------------
+ * Mutexes
+ *------------------------------------------------*/
+
+golioth_sys_mutex_t golioth_sys_mutex_create(void)
+{
+    pthread_mutex_t *mutex = golioth_sys_malloc(sizeof(pthread_mutex_t));
+    if (!mutex)
+    {
+        GLTH_LOGE(TAG, "Failed to allocate memory for mutex");
+        return NULL;
+    }
+
+    int err = pthread_mutex_init(mutex, NULL);
+    if (err)
+    {
+        GLTH_LOGE(TAG, "Failed to initialize mutex");
+        golioth_sys_free(mutex);
+        return NULL;
+    }
+
+    return (golioth_sys_mutex_t) mutex;
+}
+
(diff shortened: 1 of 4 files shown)


port/linux/golioth_sys_linux.c
```diff
@@ -47,6 +47,54 @@ uint64_t golioth_sys_now_ms(void)
...
+
+golioth_sys_mutex_t golioth_sys_mutex_create(void)
+{
+    pthread_mutex_t *mutex = golioth_sys_malloc(sizeof(pthread_mutex_t));
+    int err = pthread_mutex_init(mutex, NULL);
+    if (err)
+    {
+        assert(false);
+        return NULL;
+    }
```
sam-golioth, line 62: We can just assert that `err == 0` directly. Also, if there is an error, we should free the mutex before returning `NULL`, or we will leak memory.

port/linux/golioth_sys_linux.c
```diff
@@ -47,6 +47,54 @@ uint64_t golioth_sys_now_ms(void)
     return now_ms - start_ms;
 }
 
+/*--------------------------------------------------
+ * Mutexes
+ *------------------------------------------------*/
+
+golioth_sys_mutex_t golioth_sys_mutex_create(void)
+{
+    pthread_mutex_t *mutex = golioth_sys_malloc(sizeof(pthread_mutex_t));
```
sam-golioth, line 56: We should check that this allocation succeeds

sam-golioth (approved):

port/linux/golioth_sys_linux.c
```diff
@@ -47,6 +47,63 @@ uint64_t golioth_sys_now_ms(void)
...
+    if (!mutex)
+    {
+        GLTH_LOGE(TAG, "Failed to allocate memory for mutex");
+        return NULL;
+    }
+
+    int err = pthread_mutex_init(mutex, NULL);
+    assert(0 == err);
```
sam-golioth, line 64: I don't think we should assert here. Mutex initialization is most likely to happen during system startup, so asserting here could put us into a boot loop. I think freeing and returning NULL (as below) is fine.

sam-golioth (approved):

PR #565, opened 2024-08-21
Format port files

szczys:
Run clang-format on esp-idf, linux, and Zephyr .c and .h files.


sam-golioth (approved):
Thanks for taking the initiative here @szczys!

PR #566, opened 2024-08-22
examples: zephyr: certificate_provisioning: support 'native_sim'

mniestroj:
Support 'native_sim' platform by configuring settings storage partition.

Instead of using mcumgr over serial (as with hardware platforms), use
mcumgr over UDP. This simplifies mcumgr configuration, as there is no
serial tty when running native_sim platform, but rather stdin and stdout
streams. While it is possible to wrap it to a pseudo-terminal, it is just
much easier (by changing configuration) to switch to UDP transport instead.


mniestroj:
Those commits are only touching `certificate_provisioning` example, which succeeds with `native_sim`. Thus this PR is ready to be reviewed.

examples/zephyr/certificate_provisioning/pytest/conftest.py
```diff
@@ -15,14 +17,26 @@ def pytest_addoption(parser):
...
-def device_port(request):
+def get_device_port(request):
     if request.config.getoption("--device-port") is not None:
         return request.config.getoption("--device-port")
     else:
         port_key = f"CI_{os.environ['hil_board'].upper()}_PORT"
         return os.environ[port_key]
 
+@pytest.fixture
+def mcumgr_conn_args(request, dut):
```
sam-golioth, line 28: This is a really elegant solution!

sam-golioth (approved):
Awesome! Love that we have this test now

szczys (approved):

mniestroj:
@sam-golioth Do we want those changes in SDK 0.15.0 or merge them after SDK release? I think we can include those changes in 0.15.0 due to being very small and mainly related to testing.

sam-golioth:
> @sam-golioth Do we want those changes in SDK 0.15.0 or merge them after SDK release? I think we can include those changes in 0.15.0 due to being very small and mainly related to testing.

Let's to ahead and include them in 0.15,0

PR #569, opened 2024-08-27
zephyr: revert NXP Ethernet driver

szczys:
Zephyr v3.7.0 moves to NXP Ethernet driver which was exhibiting Link Down/Up behavior during OTA Firmware Update that resulted in the inability to download new firmware. This behavior is similar to issues reported upstream:

- https://github.com/zephyrproject-rtos/zephyr/issues/76446
- https://github.com/zephyrproject-rtos/zephyr/issues/76965

Cherry-picking the suggested patch did not resolve the issue. This commit reverts to the old driver using an overlay file for the mimxrt1024_evk. When there is an upstream fix available, these overlay files may be removed (note the certificate_provisioning example still needs the stab node in its board overlay file).

resolves https://github.com/golioth/firmware-issue-tracker/issues/642

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Workarounds:
+
+- The MIMXRT1024-EVK received a new upstream Ethernet driver that was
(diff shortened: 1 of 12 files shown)


examples/zephyr/fw_update/config.bak
```diff
@@ -0,0 +1,2478 @@
+CONFIG_DNS_SERVER_IP_ADDRESSES=y
```
hasheddan, line 1: @szczys I believe this file was unintentionally included

mniestroj (approved):

sam-golioth (approved):

PR #570, opened 2024-09-03
coap: zephyr: implement server-negotiated block size during block upload

szczys:
- Add a `negotiated_blocksize` context member to store desired block size from server
- Test the `negotiated_blocksize` and update `blockwise_context` accordingly
- Use the `negotiated_blocksize` when appending the `block1` szx (block size) on future coap packets.
    - At this point in the code we don't ave access to `blockwise_context` so this is the next best option. This is a bit convoluted and could be cleaned up in future work.

resolves https://github.com/golioth/firmware-issue-tracker/issues/628


src/coap_blockwise.c
```diff
@@ -103,6 +99,19 @@ static int call_read_block_callback(struct blockwise_transfer *ctx)
...
+// Convert index values between block sizes
+static int recalculate_block_idx(int from_idx, size_t from_larger_szx, size_t to_smaller_szx)
+{
+    assert(from_larger_szx >= to_smaller_szx);
+
+    /* Recalculate index so what was sent is now based on the new block_size */
+    int blocks_sent = from_idx + 1;
+    int size_change_multiplier = (1 << (from_larger_szx - to_smaller_szx));
+
+    /* Subtract 1 to indicate the block we just sent; inc happens after return */
```
sam-golioth, line 111: This function shouldn't assume what the caller will or will not do. Instead, it should describe what it does, and the caller can act accordingly.

src/zephyr_coap_req.c
```diff
@@ -281,6 +282,26 @@ static int golioth_coap_req_reply_handler(struct golioth_coap_req *req,
     }
     else
     {
+        if (coap_get_option_int(response, COAP_OPTION_BLOCK1) >= 0)
+        {
+            /* This response has block1 */
+            if (coap_update_from_block(response, &req->block_ctx) == 0)
+            {
+                golioth_coap_request_msg_t *rmsg = req->user_data;
```
sam-golioth, line 290: We should not be modifying user data inside this function. The expectation of the caller is that user data is passed unmodified.

Rather trying to overload the user_data with more than one purpose, we should additional arguments to the functions in the call chain so we can pass back the negotiated block size explicitly.

sam-golioth (changes requested):

szczys:
In the latest updates I used a private set_block callback typedef in coap_client.h.

I think we should also move the get_block callback out of include/golioth/client.h and into the coap_client.h as it is only referenced by internal functions not available in our api.

https://github.com/golioth/golioth-firmware-sdk/blob/2f61304dffaf9d1b5da5a0d22d657440c4b75f38/include/golioth/client.h#L162-L181

PR #578, opened 2024-09-06
examples: zephyr: use 'sysbuild' with NCS

mniestroj:
With NCS 2.7.0 comes support for 'sysbuild'. Use it, so there is the same
behavior with vanilla Zephyr and NCS.


sam-golioth (approved):

mniestroj:
@szczys Please take a look as well. Other repositories which you maintain might benefit from this change as well (e.g. reference, documentation).

See https://github.com/golioth/docs/pull/434 for documentation update.

szczys (approved):

PR #579, opened 2024-09-06
settings: check for server error in settings callback

szczys:
When the settings observation handler runs, first check the response status and log/return errors if necessary before trying to process a payload.

resolves https://github.com/golioth/firmware-issue-tracker/issues/639

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -173,18 +173,16 @@ static int finalize_and_send_response(struct golioth_client *client,
                             response->zse->payload - response->buf,
                             GOLIOTH_DEBUG_LOG_LEVEL_DEBUG);
 
-    golioth_coap_client_set(client,
-                            SETTINGS_PATH_PREFIX,
-                            "status",
-                            GOLIOTH_CONTENT_TYPE_CBOR,
-                            response->buf,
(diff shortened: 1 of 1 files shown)


mniestroj (approved):

src/settings.c
```diff
@@ -372,6 +372,16 @@ static void on_settings(struct golioth_client *client,
                         size_t payload_size,
                         void *arg)
 {
+    if (GOLIOTH_OK != response->status)
+    {
+        GLTH_LOGE(TAG,
+                  "Settings callback received status error: %d  CoAP error: %d.%02d",
+                  response->status_code,
+                  response->status_class,
+                  response->status_code);
```
sam-golioth, line 381: We're logging `response->status_code` twice, did you mean to log `response->status`?
```suggestion
        GLTH_LOGE(TAG,
                  "Settings callback received status error: %d  CoAP error: %d.%02d",
                  response->status,
                  response->status_class,
                  response->status_code);
```

szczys: Thanks for catching this!

sam-golioth (approved):

PR #581, opened 2024-09-10
tests: hil: use async to communicate with HIL

mniestroj:
Use `trio.run_process` to handle subprocess communication with
Linux (virtual) board. This simplifies receiving logic and handling
timeouts. Additionally it gives easier control over sending data to the
target (which was very difficult or impossible with
`subprocess.Popen.communicate()` API).

Use `trio-serial` for non-virtual (real hardware) boards, so the same
API and benefit as `trio.run_process()` gives is utilized for communicating
with them.

Share single `wait_for_regex_in_line()` implementation and just provide
target-specific (serial vs process) `receive_some()` API.

### DIFF
--- a/tests/hil/scripts/pytest-hil/board.py
+++ b/tests/hil/scripts/pytest-hil/board.py
@@ -1,46 +1,82 @@
 from abc import ABC, abstractmethod
-import serial
-from time import time, sleep
+from contextlib import asynccontextmanager
 import re
 
+import trio
+from trio_serial import SerialStream
+
 class Board(ABC):
-    def __init__(self, port, baud, wifi_ssid, wifi_psk, fw_image, serial_number):
+    def __init__(self, port, baudrate, wifi_ssid, wifi_psk, fw_image, serial_number):
         if serial_number:
             self.serial_number = serial_number
         self.port = port
+        self.baudrate = baudrate
+        self.fw_image = fw_image
+        self.wifi_ssid = wifi_ssid
+        self.wifi_psk = wifi_psk
 
-        if fw_image:
-            self.program(fw_image)
+        self.received: bytes = b''
+
+        self.serial: SerialStream | None = None
+
+    @asynccontextmanager
+    async def started(self):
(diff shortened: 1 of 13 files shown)


tests/hil/scripts/pytest-hil/board.py
```diff
@@ -1,46 +1,89 @@
...
+
+        self.received: bytes = b''
+
+        self.nursery: trio.Nursery | None = None
+        self.serial: SerialStream | None = None
+
+    @asynccontextmanager
+    async def started(self):
+        async with trio.open_nursery() as nursery:
+            self.nursery = nursery
```
sam-golioth, line 28: Is the idea that we'll use `self.nursery` in a later change to do concurrent send and receive with the native sim process? Or what is the purpose of the nursery here?

tests/hil/scripts/pytest-hil/board.py
```diff
@@ -1,46 +1,84 @@
...
+    @asynccontextmanager
+    async def started(self):
+        async with SerialStream(port=self.port,
+                                baudrate=self.baudrate) as serial:
+            self.serial = serial
 
-        if fw_image:
-            self.program(fw_image)
+            if self.fw_image:
+                self.program(self.fw_image)
```
sam-golioth, line 31: I wonder if the issue with the ESP-IDF tests is that we are programming the board _after_ opening the SerialStream, since the Espressif devkits are programmed over serial.

mniestroj: This indeed made it work. Great catch!

tests/hil/scripts/pytest-hil/board.py
mniestroj: I've removed `nursery` in the HIL tests with hardware (non-virtual). It seems it is not needed for serial communications with `trio-serial`.

sam-golioth (approved):
This is much simpler and a great improvement in reusability, thanks @mniestroj !

PR #586, opened 2024-09-10
ci: use esptool.py from pip instead of hal_espressif

sam-golioth:
By default, the esp32 west runner looks in the espressif HAL for the esptool.py script. Because we don't want to checkout the HAL during twister runs in CI (to save time), instead specify the path for esptool.py so that the runner will use the version installed from pip in the Docker container.

Resolves golioth/firmware-issue-tracker#652

### DIFF
--- a/.github/workflows/hil_tests.yml
+++ b/.github/workflows/hil_tests.yml
@@ -126,7 +126,7 @@ jobs:
           - hil_board: esp32_devkitc_wrover
             west_board: esp32_devkitc_wrover/esp32/procpu
             manifest: .ci-west-zephyr.yml
-            extra_twister_args: '--flash-before'
+            extra_twister_args: '--flash-before --west-flash=--esp-tool=$(which esptool.py),--esp-device=${!PORT_VAR}'
             binary_blob: hal_espressif
             run_tests: true
 


.github/workflows/hil_tests.yml
```diff
@@ -126,7 +126,7 @@ jobs:
           - hil_board: esp32_devkitc_wrover
             west_board: esp32_devkitc_wrover/esp32/procpu
             manifest: .ci-west-zephyr.yml
-            extra_twister_args: '--flash-before'
+            extra_twister_args: '--flash-before --west-flash=--esp-tool=`which esptool.py`,--esp-device=${!PORT_VAR}'
```
mniestroj, line 129:
```suggestion
            extra_twister_args: '--flash-before --west-flash=--esp-tool=$(which esptool.py),--esp-device=${!PORT_VAR}'
```
I see `$(command)` syntax getting more and more popular compared to backticks. After using it for a while I find it more readable, easier to maintain and more future-proof.

See https://mywiki.wooledge.org/BashFAQ/082

mniestroj (approved):
LGTM, just a minor comment about bash command execution syntax.

PR #592, opened 2024-09-13
Add Allure reports to hil integration tests

szczys:
- Add Allure report fixtures to the pytest plugin used for our HIL integration testing
- Add pytest options to generate Allure report on each HIL integration workflow
- Gather reports from all workflows into a single archive. This archive uses the normal retention policy, while the individual files uses a 1 day retention as they are also present in  the single archive.

## Known Issues

- If the board fixture fails, all subsequent test cases will not be categorized by the board-name but instead be shown in Allure as the generic path to the test directory. I've opened [a separate ticket](https://github.com/golioth/firmware-issue-tracker/issues/665) to track this issue.

resolves https://github.com/golioth/firmware-issue-tracker/issues/580

### DIFF
--- a/.github/workflows/hil_tests.yml
+++ b/.github/workflows/hil_tests.yml
@@ -328,3 +328,28 @@ jobs:
       api-key-id: ${{ inputs.api-key-id }}
       coap_gateway_url: ${{ inputs.coap_gateway_url }}
     secrets: inherit
+
+  merge_reports:
+    runs-on: ubuntu-latest
+    needs:
+      - hil_sample_esp-idf
+      - hil_sample_zephyr
+      - hil_sample_zephyr_nsim
+      - hil_test_esp-idf
+      - hil_test_linux
+      - hil_test_zephyr
+
+    if: ${{ always() }}
+    steps:
+    - name: Gather reports
+      uses: actions/download-artifact@v4
+      with:
+        path: allure-reports
+        pattern: allure-reports-*
+        merge-multiple: true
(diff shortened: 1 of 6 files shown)


.github/workflows/hil_tests.yml
```diff
@@ -321,3 +321,28 @@ jobs:
...
+           ]
+    if: ${{ always() }}
+    steps:
+    - name: Gather reports
+      uses: actions/download-artifact@v4
+      with:
+        path: allure-reports
+        pattern: allure-reports-*
+        merge-multiple: true
+    - run: ls -lah allure-reports
```
szczys, line 343: This can be removed before merging

szczys, line 343:
```suggestion
```

This can be removed before merging

.github/workflows/hil_tests.yml
```diff
@@ -321,3 +321,28 @@ jobs:
...
+  merge_reports:
+    runs-on: ubuntu-latest
+    needs: [
+              hil_sample_esp-idf,
+              hil_sample_zephyr,
+              hil_sample_zephyr_nsim,
+              hil_test_esp-idf,
+              hil_test_linux,
+              hil_test_zephyr,
+           ]
```
mniestroj, line 334: How about utilizing yaml (non-json) syntax:
```
needs:
  - hil_sample_esp-idf
  - hil_sample_zephyr
  - ...
```

PR #593, opened 2024-09-13
Add pytest-zephyr-samples plugin

szczys:
This plug allows common options like WiFi credentials to be added to pytest from a central location rather than doing so in  every sample's contest.py file.

This plugin will be extended to add Allure reports to the samples in an upcoming PR.


sam-golioth (approved):
Love a PR that deletes more than it adds! Nice idea @szczys !

mniestroj (approved):

PR #594, opened 2024-09-13
Add Allure reports to Zephyr twister tests

szczys:
Add Allure reports fixture to the pytest-zephyr-samples plugin so that reports are generated for all Twister tests.

#592 adds a workflow job that gathers individual reports from each job. That  will also gather these Zephyr sample reports once it and this PR are on main.

Depends on #593 
Resolved  https://github.com/golioth/firmware-issue-tracker/issues/660

### DIFF
--- a/.github/workflows/hil_sample_zephyr.yml
+++ b/.github/workflows/hil_sample_zephyr.yml
@@ -181,6 +181,7 @@ jobs:
           hil_board: ${{ inputs.hil_board }}
           west_board: ${{ inputs.west_board }}
         run: |
+          rm -rf allure-reports
           source /opt/credentials/runner_env.sh
           export PATH=$PATH:/opt/SEGGER/JLink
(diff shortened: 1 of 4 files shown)


sam-golioth (approved):

.github/workflows/hil_sample_zephyr.yml
```diff
@@ -160,6 +160,7 @@ jobs:
           git config --global user.name "Git User"
           west patch
 
+          pip3 install allure-pytest
```
mniestroj, line 163: Is it still needed after adding `allure-pytest` to [pyproject.toml](https://github.com/golioth/golioth-firmware-sdk/pull/594/commits/e1d626ff18e78cef0d8814699258d191a23f4b43#diff-351b54743528ead0e201e113dd12d1392d22392d8b838ea7fe3a8f35ada115c7)?

mniestroj (approved):

PR #597, opened 2024-09-15
workflows: mask secrets in logs before upload

szczys:


sam-golioth (approved):
This is great @szczys! Thanks for taking the initiative to correct this and well done finding a quick and elegant solution.

We should merge this as is, but I imagine we're not the only one facing this or a similar problem. We should consider publishing this action in its own repository.

mniestroj (approved):

szczys:
The secrets were not masked on the native_sim runs. It looks like it uses a specific commit of the docker images:
https://github.com/golioth/golioth-firmware-sdk/blob/ef4600978b64cfdef71246e6d960ca6c67648b17/.github/workflows/hil_tests.yml#L178

I will need to update that image to include jq (or perhaps add a conditional check to install it using apt).

PR #601, opened 2024-09-18
Add a safe-upload-artifacts action

szczys:
This combined the functionality of the mask_secrets and upload-artifact actions.

This action takes a `name` and `path` input in the same formatting as the `upload-artifact` action, masks any GitHub actions found in the files on those paths, and then calls the upload-artifact action to upload them as expected.

resolves https://github.com/golioth/firmware-issue-tracker/issues/669


.github/actions/safe-upload-artifacts/action.yml
```diff
@@ -0,0 +1,71 @@
...
+
+        for key in $(jq -r "keys[]" <<< "$SECRETS_CONTEXT");
+        do
+          secret_val=$(jq -r ".$key" <<< "$SECRETS_CONTEXT")
+
+          if [[ ! $secret_val =~ "\n" ]]; then
+            # This approach to escaping the regex found: https://stackoverflow.com/a/29613573/922013
+            ESCAPED_SECRET=$(sed 's/[^^]/[&]/g; s/\^/\\^/g' <<< "$secret_val")
+
+            # Iterate list of input paths and use find to create a file list from the patterns
```
szczys, line 51:
```suggestion
            # Iterate list of input path patterns and use grep to create a list of files that contain secrets
```

.github/actions/safe-upload-artifacts/action.yml
```diff
@@ -0,0 +1,71 @@
...
+            while IFS= read -r search_path || [[ -n $search_path ]]; do
+              [ $(grep -Rl $ESCAPED_SECRET $search_path 2>/dev/null > __grep_search_output.txt) >= 0 ]
+            done < <(printf '%s' "${{ inputs.path }}")
+
+            uniq __grep_search_output.txt > __files_to_mask.txt
+
+            if [ -s __grep_search_output.txt ]; then
```
szczys, line 58:
```suggestion
            if [ -s __files_to_mask.txt ]; then
```

.github/actions/safe-upload-artifacts/action.yml
```diff
@@ -0,0 +1,73 @@
+name: Mask secrets in files
+
+description: |
+  Search all files in a give path(s) and replace any GitHub secrets with ***NAME_OF_SECRET***
```
sam-golioth, line 4:
```suggestion
  Search all files in a given path(s) and replace any GitHub secrets with ***NAME_OF_SECRET***
```

.github/actions/safe-upload-artifacts/action.yml
```diff
@@ -0,0 +1,73 @@
...
+            while IFS= read -r search_path || [[ -n $search_path ]];
+            do
+              [ $(grep -Rl $ESCAPED_SECRET $search_path 2>/dev/null >> __grep_search_output.txt) >= 0 ]
+            done < <(printf '%s' "${{ inputs.path }}")
+
+            uniq __grep_search_output.txt > __files_to_mask.txt
+
+            if [ -s __files_to_mask.txt ]; then
+              cat __files_to_mask.txt | xargs -I{} sed -i "s/$ESCAPED_SECRET/***$key***/g" {}
+            fi
```
sam-golioth, line 62: We can remove the need to write another file:
```suggestion
            if [ -s __grep_search_output.txt ]; then
              uniq __grep_search_output.txt | xargs -I{} sed -i "s/$ESCAPED_SECRET/***$key***/g" {}
            fi
```

.github/actions/safe-upload-artifacts/action.yml
```diff
@@ -0,0 +1,73 @@
...
+
+            rm -rf __grep_search_output.txt __files_to_mask.txt
+          fi
+        done
+
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v4
+      if: steps.find-and-mask.outcome == 'success'
```
sam-golioth, line 70: I think this is implied so we don't require it. We needed this before because we were using an `if: always()` directive, but now that's applied for the whole action instead of just the individual upload artifacts step.

sam-golioth (approved):
Very cool!

PR #603, opened 2024-09-18
ci: add timeout for self-hosted runners

sam-golioth:
A timeout will ensure that a hung test does prevent other tests that need to use self-hosted runners from running.


szczys (approved):
Hallelujah!

PR #610, opened 2024-09-19
ci: coverage: generate HTML coverage

mniestroj:
This generates HTML coverage from all native_sim runs:
 * both 32-bit and 64-bit
 * samples and tests
 
Echo HIL and Twister gcovr tracefiles, which helps with visibility of what is actually combined to single gcovr output.

Drop HTML coverage generation from specific native_sim jobs, as that just shows partial result.

### DIFF
--- a/.github/workflows/hil_tests.yml
+++ b/.github/workflows/hil_tests.yml
@@ -314,7 +314,7 @@ jobs:
               --platform ${{ matrix.west_board }}                                   \
               -T modules/lib/golioth-firmware-sdk/examples/zephyr                   \
               -C --coverage-basedir modules/lib/golioth-firmware-sdk                \
-              --coverage-formats txt,html                                           \
+              --coverage-formats txt                                                \
               -x=CONFIG_GOLIOTH_COAP_HOST_URI=\"${{ inputs.coap_gateway_url }}\"    \
               --pytest-args="--api-url=${{ inputs.api-url }}"                       \
               --pytest-args="--api-key=${{ secrets[inputs.api-key-id] }}"           \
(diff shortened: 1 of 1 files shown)


mniestroj:
This PR addresses HTML coverage generation alone. Deployment to Firebase (or any other hosting) will be done in separate PR.

.github/workflows/hil_tests.yml
```diff
@@ -407,7 +410,17 @@ jobs:
             $hil_tracefiles $twister_tracefiles \
             -e examples -e external -e tests"
 
-          gcovr --txt-summary --xml coverage.xml
+          rm -rf coverage
+          mkdir -p coverage
+          gcovr --txt-summary --xml coverage.xml --html-details coverage/index.html
+
+      - name: Upload coverage artifacts
+        uses: actions/upload-artifact@v4
```
szczys, line 422: Can we use safe-upload-artifacts here?

szczys (approved):

sam-golioth (approved):

.github/workflows/hil_tests.yml
mniestroj: Sure we can... but should we? There are just HTML generated pages, with code and colors applied based on coverage. There are no logs there, so no secrets. IMO we should use generic stuff for things that do not require special treatment.

sam-golioth: I'm also hesitant on this one in particular because we could (I don't think likely, but possible) inadvertently change the source code that's displayed in the HTML report which could make searching or comparing more difficult.

PR #612, opened 2024-09-20
coap: zephyr: add log level zephyr coap client

szczys:
The coap_client_zephyr.c file registers for logging but it does not pass a log level. This meant that enabling debug logging for the Golioth Firmware SDK did not include enabling logs from the zephyr client. This commit adds GOLIOTH_DEBUG_DEFAULT_LOG_LEVEL as the default log level.

resolves https://github.com/golioth/firmware-issue-tracker/issues/598

### DIFF
--- a/src/coap_client_zephyr.c
+++ b/src/coap_client_zephyr.c
@@ -24,7 +24,7 @@
 #include <golioth_ciphersuites.h>
 #include "golioth_openthread.h"
 
-LOG_MODULE_REGISTER(golioth_coap_client_zephyr);
+LOG_TAG_DEFINE(golioth_coap_client_zephyr);
 
 #define GOLIOTH_MAX_IDENTITY_LEN 32
 #define GOLIOTH_EMPTY_PACKET_LEN (16 + GOLIOTH_MAX_IDENTITY_LEN)


mniestroj (approved):

src/coap_client_zephyr.c
```diff
@@ -24,7 +24,7 @@
 #include <golioth_ciphersuites.h>
 #include "golioth_openthread.h"
 
-LOG_MODULE_REGISTER(golioth_coap_client_zephyr);
+LOG_MODULE_REGISTER(golioth_coap_client_zephyr, CONFIG_GOLIOTH_DEBUG_DEFAULT_LOG_LEVEL);
```
hasheddan, line 27:
```suggestion
LOG_TAG_DEFINE(golioth_coap_client_zephyr);
```

I think we should use `LOG_TAG_DEFINE` here to be consistent with other `src` files, despite this being Zephyr only. `LOG_TAG_DEFINE` is [defined in the Zephyr port](https://github.com/golioth/golioth-firmware-sdk/blob/3e959c2f6fd96b560ee9791c3527ec7f8f073db2/port/zephyr/include/golioth_port_config.h#L11) as `LOG_MODULE_REGISTER(tag, CONFIG_GOLIOTH_LOG_LEVEL)`, so using `LOG_MODULE_REGISTER(golioth_coap_client_zephyr, CONFIG_GOLIOTH_DEBUG_DEFAULT_LOG_LEVEL)` here results in this one module's logging being controlled by a different config (`CONFIG_GOLIOTH_DEBUG_DEFAULT_LOG_LEVEL`) than the others (`CONFIG_GOLIOTH_LOG_LEVEL`), which is somewhat confusing. Also, if there was a change made in the Zephyr port as to how `LOG_TAG_DEFINE` behaves in the future, it would not propagate here if we are using `LOG_MODULE_REGISTER` directly.

Longer term I think it would be good to use a single config for log level control across ports. I believe the reason for the discrepancy today is because we are using the Zephyr [Kconfig logging template](https://github.com/golioth/golioth-firmware-sdk/blob/3e959c2f6fd96b560ee9791c3527ec7f8f073db2/port/zephyr/Kconfig#L189).

hasheddan (approved):

sam-golioth (approved):

PR #618, opened 2024-09-23
Add rak5010 to integration tests

szczys:
Resolves https://github.com/golioth/firmware-issue-tracker/issues/653


tests/hil/platform/zephyr/boards/rak5010_nrf52840.overlay
```diff
@@ -0,0 +1,13 @@
...
+    zephyr,shell-uart = &cdc_acm_uart0;
+  };
+
+};
+
+&zephyr_udc0 {
+  cdc_acm_uart0: cdc_acm_uart0 {
+    compatible = "zephyr,cdc-acm-uart";
+  };
+};
```
mniestroj, line 13: Please use TABs here for identation.

tests/hil/scripts/pytest-hil/rak5010.py
```diff
@@ -0,0 +1,27 @@
+from zephyrboard import ZephyrBoard
+import subprocess
+from time import sleep
+
+class RAK5010(ZephyrBoard):
+    def __init__(self, *args, **kwargs):
+        ZephyrBoard.__init__(self, *args, **kwargs)
```
mniestroj, line 7: This can be removed, as it does nothing `RAK5010` specific.

tests/hil/scripts/pytest-hil/rak5010.py
```diff
@@ -0,0 +1,27 @@
...
+    def program(self, fw_image):
+        command = ['gdb',
+                   '-ex', "set confirm off",
+                   '-ex', "target extended-remote {}".format(self.bmp_port),
+                   '-ex', "monitor swdp_scan",
+                   '-ex', "attach 1",
+                   '-ex', f"load {fw_image}",
+                   '-ex', "kill",
+                   '-ex', "quit"
+                   ]
```
mniestroj, line 18: This is not `RAK5010` platform- specific. How about naming this class `BMPBoard` instead of `RAK5010`?

mniestroj (approved):

tests/hil/scripts/pytest-hil/bmpboard.py
```diff
@@ -0,0 +1,24 @@
...
+                   '-ex', "quit"
+                   ]
+
+        subprocess.run(command)
+
+        # Arbitrary sleep while the software-based USB serial becomes available
+        sleep(3)
+
+    @property
+    def USES_WIFI(self):
```
sam-golioth, line 23: Rather than assume this for all boards that use the black magic probe, we should have an additional class for the RAK5010 that inherits from both `ZephyrBoard` and `BmpBoard`. See the 1024 for an example (inheriting from `ZephyrBoard` and `JLinkBoard`).

tests/hil/scripts/pytest-hil/bmpboard.py
```diff
@@ -0,0 +1,24 @@
+from zephyrboard import ZephyrBoard
+import subprocess
+from time import sleep
+
+class BmpBoard(ZephyrBoard):
```
sam-golioth, line 5: NIT: I think we should capitalize BMP: `BMPBoard`.

tests/hil/scripts/pytest-hil/bmpboard.py
mniestroj: Okay, so maybe let's move back to `RAK5010` and only introduce `BMPBoard` (or `BmpBoard`) once needed for other platforms. `sleep(3)` above in `__init__()` is also RAK5010 specific.

sam-golioth: I'm ok either way, I think it could be more readable to add the additional class (to make it clear what's bmp-specific and what's RAK5010 specific). We can add the delay by calling the super function:

```
def program(self, *args, **kwargs):
    BMPBoard.program(self, *args, **kwargs)
    
    sleep(3)
```

szczys: Good suggestions, I've broken this out into the separate programmer and board files.

tests/hil/scripts/pytest-hil/nrf52840dk.py
```diff
@@ -2,6 +2,12 @@
 from nordicboard import NordicBoard
 
 class nRF52840DK(ZephyrBoard, NordicBoard):
+    def program(self, *args, **kwargs):
```
sam-golioth, line 5: Wrong class?

szczys: Indeed, sorry about that. Rerunning tests after correcting this.

sam-golioth (approved):
Filling in the puzzle!

PR #627, opened 2024-09-26
ci: linux: switch from lcov to gcovr

mniestroj:
Zephyr tests and samples coverage is handled using gcovr. Switch to that
tool for Linux tests as well for improved consistency. This also allows to
combine coverage results for Linux and Zephyr.

Use gcovr tool to generate JSON output for each test and put most of the
coverage report generation inside 'coverage' job:
 * HTML report (instead of lcov -> genhtml in case of Linux)
 * single PR comment from XML output (instead of having two comments, each
   with slightly different format)
 * GH Actions step summary that includes Linux platform coverage (which was
   not generated before for Linux)

### DIFF
--- a/.github/workflows/hil_test_linux.yml
+++ b/.github/workflows/hil_test_linux.yml
@@ -47,25 +47,20 @@ jobs:
     - name: Setup Python dependencies
       run: |
         pip install --upgrade pip
-        pip install pytest pytest-timeout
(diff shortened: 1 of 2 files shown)


mniestroj:
> The coverage check is failing on this PR.

Fixed that now! Coverage artifacts were not downloading into correct directory.

sam-golioth (approved):

PR #631, opened 2024-09-27
Block download: add current_block_len

szczys:
Block download callbacks now receive both block_size and current_block_len. These distinct values are needed to correctly calculate the offset when the final block is received and its data payload (current_data_len) is smaller than the coap blocksize (block_size).

Resolves https://github.com/golioth/firmware-issue-tracker/issues/656

### DIFF
--- a/src/coap_blockwise.h
+++ b/src/coap_blockwise.h
@@ -28,8 +28,9 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
 /* Blockwise Download */
 typedef enum golioth_status (*write_block_cb)(uint32_t block_idx,
                                               uint8_t *block_buffer,
-                                              size_t block_size,
+                                              size_t block_buffer_len,
                                               bool is_last,
+                                              size_t negotiated_block_size,
                                               void *callback_arg);
 
 enum golioth_status golioth_blockwise_get(struct golioth_client *client,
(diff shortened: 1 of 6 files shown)


src/coap_blockwise.h
```diff
@@ -27,8 +27,9 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
 
 /* Blockwise Download */
 typedef enum golioth_status (*write_block_cb)(uint32_t block_idx,
-                                              uint8_t *block_buffer,
                                               size_t block_size,
+                                              uint8_t *block_buffer,
+                                              size_t current_block_len,
                                               bool is_last,
                                               void *callback_arg);
```
sam-golioth, line 34: I think we should rearrange/rename some of these:
- I think it makes sense to keep all the fields relating to the current block together (`block_idx`, `block_buffer`, `current_block_len`, and `is_last`). So I would move `block_size` to just before `callback_arg`.
- It will be clearer if we keep the buffer and length naming consistent - so instead of `current_block_len`, we could use `block_buffer_len`.
- Finally, to further remove ambiguity, renaming `block_size` to `negotiated_block_size` should make things crystal clear.

sam-golioth (commented):
Code changes look good, just left one note on argument naming. Also, we should squash these commits into one - the first commit compiles, but it won't run correctly because the meaning of `block_size` has changed.

sam-golioth (approved):
Looks great, thanks @szczys!

PR #636, opened 2024-10-01
ci: unit_tests: merge coverage with HIL tests

mniestroj:
`tests/unit_tests/test.sh` is very verbose, while it is possible to do most
of it with single `ctest` command.

Use *Ninja* generator to build using multiple threads automatically.

Install both `gcovr` and `ninja` from pypi, so a single tool (pip) is used.

Move `unit_tests` job from `lint_build_unit_test.yml` to `hil_tests.yml`,
so that coverage data can be easily combined as part of single workflow.

Also rename `lint_build_unit_test.yml` to `lint_build.yml`, as there are no
more unit tests as part of this workflow.

### DIFF
--- a/.github/workflows/hil_tests.yml
+++ b/.github/workflows/hil_tests.yml
@@ -53,6 +53,43 @@ on:
         default: 'all'
 
 jobs:
+  unit_tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository and submodules
+        uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
+
+      - name: Install gcovr and ninja
+        run: |
+          pip install gcovr ninja
+
+      - name: Run unit tests
+        run: |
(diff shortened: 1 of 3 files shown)


szczys (approved):

sam-golioth (approved):
There's a lot of little improvements here that have been bugging me for a while! I wasn't a fan of calling out to another shell script for testing, I wonder if we can just delete that whole script now? I also wonder if we should rename `hil_tests.yml` to just `test.yml` since we have a few tests in there now which don't run on hardware.

mniestroj:
> There's a lot of little improvements here that have been bugging me for a while! I wasn't a fan of calling out to another shell script for testing, I wonder if we can just delete that whole script now? I also wonder if we should rename `hil_tests.yml` to just `test.yml` since we have a few tests in there now which don't run on hardware.

Good suggestions, I'll create followup PRs with those changes!

PR #638, opened 2024-10-02
zephyr: logging: fix off-by-one in log message length limit

szczys:
Fix an off-by-one error when writing a null terminator to the end of a log message buffer before sending to Golioth.

When a log message is longer than the setting for the  maximum Golioth backend log buffer (CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE), the message is truncated by writing a null terminator onto the end of the buffer. The index for this null terminator was using a counter that had already been incremented past the end of the buffer array. Instead, use a MIN() test to use either the counter value, or the final member of the buffer array.

### DIFF
--- a/port/zephyr/golioth_log_zephyr.c
+++ b/port/zephyr/golioth_log_zephyr.c
@@ -72,7 +72,7 @@ static const glth_log_fn log_to_log_func(struct log_msg *log)
 static int cbpprintf_out_func(int c, void *out_ctx)
 {
     struct cbpprintf_ctx *ctx = out_ctx;
-    if (ctx->ctr >= CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE)
+    if (ctx->ctr >= CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE - 1)
     {
         return -ENOMEM;
     }


mniestroj (approved):

port/zephyr/golioth_log_zephyr.c
```diff
@@ -106,7 +106,10 @@ static void process(const struct log_backend *const backend, union log_msg_gener
     void *data = log_msg_get_package(log, &len);
     ctx->print_ctx.ctr = 0;
     cbpprintf(cbpprintf_out_func, &ctx->print_ctx, data);
-    ctx->print_ctx.msg[ctx->print_ctx.ctr] = '\0';
+
+    int null_term_idx = MIN(ctx->print_ctx.ctr, CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE - 1);
+    ctx->print_ctx.msg[null_term_idx] = '\0';
+
```
mniestroj, line 112: Shouldn't it be enough to change:
```
 if (ctx->ctr >= CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE)
```
into
```
 if (ctx->ctr > CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE)
```
in `cbpprintf_out_func()`?

szczys: I thought about doing something like this:

```
if (ctx->ctr >= CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE - 1)
```

But that is not actually causing the problem. The problem is that before I changed line 109, it is using an index that is changed somewhere else. I think it makes sense to bound it here where it is used. This way, no matter what changes in the future, this will always be a valid index.

mniestroj: There is no other place than `cbpprintf_out_func()` that is modifying `ctr`... allowing this function to write the last byte does not make sense, since it will be overwritten by `\0` here at line 109.

Maybe it should be:
```
if (ctx->ctr > CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE - 1)
```
and in that case `MIN()` usage at line 109 won't be needed.

mniestroj (changes requested):
Blocking for now just to get an agreement in https://github.com/golioth/golioth-firmware-sdk/pull/638/commits/76872cf46db8703aa90486957bc11e2526c188ee#r1784729939.

port/zephyr/golioth_log_zephyr.c
szczys: I guess would be in favor of including both of these changes. To me it seems like a good idea to have an explicit limit when indexing an array. It was hard to track down this problem because the change was taking place in a different function from where the array write was happening.

 @sam-golioth do you have a preference on this?

mniestroj: IMO we should have one place in the implementation to enforce the limit. Having two places that guard buffer overflow is not very future-proof. Maybe a comment in `cbpprintf_out_func()` (before `if (ctx->ctr >= CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE - 1)`) about leaving one byte at the end for `NULL` terminator would be good to add?

> It was hard to track down this problem because the change was taking place in a different function from where the array write was happening.

Agree, that this kind of problem is not trivial to find. However... I still think that we should have one place to guard for buffer overflow, not two. Doing something "just in case" isn't putting too much confidence in other parts of the implementation.

mniestroj: Yep, let's give @sam-golioth the final call.

sam-golioth: I don't have a strong preference between the two proposed changes, but I agree that putting in both fixes is less readable because it makes the reader question why there exists a codepath that cannot be executed.

Checking for `CONFIG_LOG_BACKEND_GOLIOTH_MAX_LOG_STRING_SIZE - 1` in the cbprintf callback saves us a couple CPU cycles by avoiding an unnecessary write to the last byte and removing a comparison, so let's go with that one.

mniestroj (approved):

PR #641, opened 2024-10-03
Hil split

mniestroj:


sam-golioth:
Can we get some rough numbers on how much time this adds for a complete test run (ignoring retries)?  Let's also identify areas where we can cut down on the per-test overhead. Some of the ones that come to mind (but there could be others):
- erasing the board after each run
- Setting up python/west dependencies

mniestroj:
> Can we get some rough numbers on how much time this adds for a complete test run (ignoring retries)?

Number directly calculated from CI, based on `succeeded on Oct 10 in 2m 30s` strings reported by GitHub (e.g. here https://github.com/golioth/golioth-firmware-sdk/actions/runs/11274763684/job/31359278655?pr=641):

|branch|time (s)|time adjusted (s)|% (adjusted)|link|
|-|-|-|-|-|
|`hil-split`|1343|1089|170|https://github.com/golioth/golioth-firmware-sdk/actions/runs/11274763684/job/31359278655?pr=641|
|`main`|642|642|100|https://github.com/golioth/golioth-firmware-sdk/actions/runs/11284048073/job/31384742968|

Adjusted time was based on `sams_orange_pi` taking around 120s to install Python dependencies, while `tims_orange_pi` taking only 33s. Also there was one test executed on `tims_orange_pi` that resulted in long container bring-up (probably cold cache).

mniestroj:
There is one nice comparison now to make. Right now there is an issue with nrf9160dk with `ota` HIL test, which is a nice use case to describe.

This OTA test takes 3min 49s in total, on `hil-split` branch: https://github.com/golioth/golioth-firmware-sdk/actions/runs/11953157487/job/33320699596?pr=641

All tests (including failing OTA test) take 20min 42s, on another branch: https://github.com/golioth/golioth-firmware-sdk/actions/runs/11942272439/job/33326973074?pr=682

This means that restart will be much faster in execution. This means less time spent on blocking other GitHub Actions invocations.

Another important feature is that with the first link (`hil-split` branch) it is much easier to figure out which test has failed (OTA), whereas with the second link it is not that easy, as whole step log needs to be analyzed in order to see that only OTA test has failed there.

PR #644, opened 2024-10-03
Yaml fix indentation

mniestroj:


.yamllint
```diff
@@ -0,0 +1,26 @@
...
+ignore: |
+  /external
+  /docs/doxygen/doxygen-awesome-css
+  mtb_shared
+
+rules:
+  line-length:
+    max: 100
+    ignore: |
+      /.github/actions/safe-upload-artifacts/action.yml
```
sam-golioth, line 14: Why ignore these files?

mniestroj: There is also https://yamllint.readthedocs.io/en/stable/disable_with_comments.html which we could use instead of ignoring whole file.

mniestroj: Rechecked and seems that those issues with >100 columns could be solved in a reasonable way.

sam-golioth (approved):

PR #645, opened 2024-10-03
ci: rename 'hil_tests' workflow to 'tests'

mniestroj:
This workflow includes unit tests, so let's rename it.


sam-golioth:
Thanks @mniestroj! Unfortunately this will require changing all of the required checks in the repo settings to match the new name

PR #650, opened 2024-10-07
coap_blockwise: support passing set_cb to golioth_blockwise_post()

mniestroj:
`golioth_blockwise_post()` was just returning `enum golioth_status`. Add
support for `set_cb`, which will be called in case of response to blockwise
POST request.

### DIFF
--- a/src/coap_blockwise.h
+++ b/src/coap_blockwise.h
@@ -23,6 +23,7 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
                                            const char *path,
                                            enum golioth_content_type content_type,
                                            read_block_cb cb,
+                                           golioth_set_cb_fn callback,
                                            void *callback_arg);
 
 /* Blockwise Download */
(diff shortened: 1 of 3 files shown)


szczys (approved):

src/coap_blockwise.h
```diff
@@ -23,6 +23,7 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
                                            const char *path,
                                            enum golioth_content_type content_type,
                                            read_block_cb cb,
+                                           golioth_set_cb_fn callback,
```
sam-golioth, line 26: Can we add a comment explaining that this function is called once with the response to the last block? I think that users could reasonably assume (without reading the implementation), that this callback is called once for _each_ block.

sam-golioth (approved):

src/coap_blockwise.h
mniestroj: Good idea. However we don't have any documentation for that module yet. So far we kind of rely on reading the implementation, since we don't provide any documentation for blockwise trasnfer module (which is private BTW). I'll leave that as an improvement for future for subsequent PRs.

PR #652, opened 2024-10-07
workflows: zephyr: remove extra comma from runs-on entry

szczys:
Comma after board label is causing tests to not be able to find runners.


sam-golioth (approved):

sam-golioth:
Pushed this one through

PR #653, opened 2024-10-09
OTA: add function for resuming component download

szczys:
Add `golioth_ota_download_component_resumable()` that takes a block_idx to start the download and returns the next expected block_idx which may be used to resume when a block download fails before the final block.

Resolves https://github.com/golioth/firmware-issue-tracker/issues/684
Resolves https://github.com/golioth/firmware-issue-tracker/issues/686

### DIFF
--- a/include/golioth/ota.h
+++ b/include/golioth/ota.h
@@ -156,22 +156,35 @@ typedef enum golioth_status (*ota_component_block_write_cb)(
     size_t negotiated_block_size,
     void *arg);
 
-/// Download an OTA component synchronously.
+/// Begin or resume downloading an OTA component synchronously.
 ///
 /// This function will block until the final block of the component download is received from the
-/// server or an error is received on any block.
+/// server or an error is received on any block. The `next_block_idx` is used to pass in the first
+/// block index to be downloaded (components begin with the 0th index). After returning, the same
+/// parameter has been updated to indicate the next expected block in the download process.
+///
+/// Use this function to resume OTA component downloads. Pass the `next_block_idx` from a failed run
+/// to resume the download. The `ota_component_block_write_cb` function must be designed to handle
+/// storage when resuming a download. Upon successful completion, it is up to the application to
+/// verify the integrity of the component (eg: compare to the SHA256 in `struct
+/// golioth_ota_component`).
 ///
 /// @param client The client handle from @ref golioth_client_create
 /// @param component One @ref golioth_ota_component instance present in the @ref
 /// golioth_ota_manifest
+/// @param[in,out] block_idx Pointer to the next expected block index. The function will
+/// request the block index supplied as an input, when returning, this variable indicates the next
+/// block index expected. When an error status is returned, use this value as the input for the next
+/// resume operation. May be NULL to begin download with block_idx 0.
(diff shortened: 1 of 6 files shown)


sam-golioth:
What do you think about combining the two functions (in both the OTA service and in `coap_blockwise.c`), making the `block_idx` argument optional?

src/ota.c
```diff
@@ -367,6 +368,7 @@ enum golioth_status golioth_ota_download_component(struct golioth_client *client
                                  "",
                                  component->uri,
                                  GOLIOTH_CONTENT_TYPE_OCTET_STREAM,
+                                 NULL,
```
sam-golioth, line 371:
```suggestion
                                next_block_idx,
```

include/golioth/ota.h
```diff
@@ -156,22 +156,34 @@ typedef enum golioth_status (*ota_component_block_write_cb)(
...
 /// @param cb Callback function to register
 /// @param arg Optional argument, forwarded directly to the callback when invoked. Can be NULL.
 ///
-/// @retval GOLIOTH_OK all blocks of package were received
+/// @retval GOLIOTH_OK the final block of package was received
 /// @retval GOLIOTH_ERR_FAIL invalid client handle, path, or callback
 /// @retval GOLIOTH_ERR_MEM_ALLOC unable to allocate necessary memory
 enum golioth_status golioth_ota_download_component(struct golioth_client *client,
                                                    const struct golioth_ota_component *component,
+                                                   uint32_t *next_block_idx,
```
sam-golioth, line 186:
```suggestion
                                                   uint32_t *block_idx,
```
Maybe just `block_idx` here? I think the consistency with `golioth_blockwise_get()` will make it a bit easier to follow the code

include/golioth/ota.h
```diff
@@ -156,22 +156,34 @@ typedef enum golioth_status (*ota_component_block_write_cb)(
...
 /// This function will block until the final block of the component download is received from the
-/// server or an error is received on any block.
+/// server or an error is received on any block. The `next_block_idx` is used to pass in the first
+/// block index to be downloaded (components begin with the 0th index). After returning, the same
+/// parameter has been updated to indicate the next expected block in the download process.
+///
+/// Use this function to resume OTA component downloads. Pass the `next_block_idx` from a failed run
+/// to resume the download. The `ota_component_block_write_cb` function must be designed to handle
+/// storage when resuming a download. Upon successful completion, it is up to the application to
+/// verify the integrity of the component (eg: compare to the SHA256 in `golioth_ota_component`).
```
sam-golioth, line 169:
```suggestion
/// verify the integrity of the component (eg: compare to the SHA256 in `struct golioth_ota_component`).
```

sam-golioth (changes requested):
Looks like a copy/paste bug in `golioth_ota_download_component()`, otherwise looks great! I love how small this changeset is.

src/ota.c
szczys: Yeah, that's a rebase error. It's correct on this commit, but should be changed on the next commit when the param is added to the parent function. Thanks for catching!

sam-golioth (approved):

PR #654, opened 2024-10-10
ci: add codecov report

mniestroj:


mniestroj:
No plans to work on it anytime soon. In order to make use of Codecov test reporting, we would need to better extract information from our test runs and provide Codecov all data that it can parse / render. Not sure how useful it would be before implementing it. As we have fully open source and hosted tests results already implemented, it might not provide enough value to justify the effort. Closing for now.

PR #655, opened 2024-10-10
workflows: allure: separate main branch from others

szczys:
- Use separate subdirectories for building Allure reports on `main` from those built on all other branches.
- Copy the index.html for the main branch to root so the landing page always shows the latest `main` branch reports
- Update the summary link to include the subdirectory

Resolves https://github.com/golioth/firmware-issue-tracker/issues/664

### DIFF
--- a/.github/workflows/reports-allure-publish.yml
+++ b/.github/workflows/reports-allure-publish.yml
@@ -2,6 +2,10 @@ name: Publish Allure Reports to GitHub Pages
 
 on:
   workflow_call:
+  push:
+
+env:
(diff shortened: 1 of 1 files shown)


szczys:
Can we ignore [the linter error on this one](https://github.com/golioth/golioth-firmware-sdk/actions/runs/11280263352/job/31372837520#step:3:6)? There's an extra long  line when building a URL using several variables but I'd rather not break it over lines unless we really need to do so.

sam-golioth (approved):

PR #659, opened 2024-10-15
Blockwise upload error handling

szczys:
Add error handling to the blockwise upload operation.

- Propagate any error returned by the application write block callback back to the initial call.
- Add an error condition for invalid block size and check the block size returned by the application callback for bounding.
- Check the status and response at the end of a block-write operation and only call the set_cb callback if there was no local error (it should still be called for successful block upload, and errors returned from the server during coap operations).

Resolves https://github.com/golioth/firmware-issue-tracker/issues/712

This implements the parts of an existing PR that are still relevant to the current state of the SDK so this PR also:
- Closes #489

### DIFF
--- a/src/coap_blockwise.c
+++ b/src/coap_blockwise.c
@@ -90,23 +90,32 @@ static void on_block_sent(struct golioth_client *client,
     golioth_sys_sem_give(ctx->sem);
 }
 
-// Function to call the application's read block callback after a successful
-// blockwise upload
-static int call_read_block_callback(struct blockwise_transfer *ctx)
+// Function to call the application's read block callback for obtaining
+// blockwise upload data
+static enum golioth_status call_read_block_callback(struct blockwise_transfer *ctx,
+                                                    size_t *block_buffer_len)
 {
     /* Do not allow user callback to directly change ctx->block_size value */
     size_t block_size = ctx->block_size;
 
-    int err = ctx->callback.read_cb(ctx->block_idx,
-                                    ctx->block_buffer,
-                                    &block_size,
-                                    &ctx->is_last,
-                                    ctx->callback_arg);
-    if (err != GOLIOTH_OK)
+    enum golioth_status status = ctx->callback.read_cb(ctx->block_idx,
+                                                       ctx->block_buffer,
+                                                       &block_size,
+                                                       &ctx->is_last,
+                                                       ctx->callback_arg);
+    if (status != GOLIOTH_OK)
     {
-        return -err;
+        return status;
     }
-    return block_size;
+
+    if ((ctx->is_last == false && block_size != ctx->block_size) || block_size > ctx->block_size
+        || block_size == 0)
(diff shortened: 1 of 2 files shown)


src/coap_blockwise.c
```diff
@@ -90,23 +90,32 @@ static void on_block_sent(struct golioth_client *client,
...
+                                                       &ctx->is_last,
+                                                       ctx->callback_arg);
+    if (status != GOLIOTH_OK)
     {
-        return -err;
+        return status;
+    }
+
+    if ((ctx->is_last == false && block_size != ctx->block_size) || block_size > ctx->block_size
+        || block_size == 0)
```
sam-golioth, line 112: Since we're breaking this line anyway, I think it's clearer to have each condition on a separate line, rather than some combined and some separate. However, clang-format may complain about my suggestion, in which case we will respect the computer overlord.
```suggestion
    if ((ctx->is_last == false && block_size != ctx->block_size)
        || block_size > ctx->block_size
        || block_size == 0)
```

src/coap_blockwise.c
```diff
@@ -171,35 +180,20 @@ static enum golioth_status upload_single_block(struct golioth_client *client,
...
-    else
+    size_t block_buffer_len = ctx->block_size;
+    enum golioth_status status = call_read_block_callback(ctx, &block_buffer_len);
+    if (status == GOLIOTH_OK)
     {
-        status = upload_single_block(client, ctx, (size_t) ret);
+        status = upload_single_block(client, ctx, block_buffer_len);
         if (status == GOLIOTH_OK)
         {
+            /* Only advance block_idx if block was stored successfully */
```
sam-golioth, line 194:
```suggestion
            /* Only advance block_idx if block was uploaded successfully */
```

src/coap_blockwise.c
```diff
@@ -251,7 +245,13 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
 
     if (set_cb)
     {
-        set_cb(client, &ctx->response, path, callback_arg);
+        bool local_error =
+            (status != GOLIOTH_OK && (ctx->block_idx == 0 || ctx->response.status == GOLIOTH_OK));
```
sam-golioth, line 249: I'm a little worried about this being brittle, since it's not obvious that the response from the previous block should remain useable while processing the next block. I think just adding a comment explaining this boolean logic should suffice for now.

sam-golioth (approved):

src/coap_blockwise.c
szczys: I agree with you and [I had it formatted the same way](https://github.com/golioth/golioth-firmware-sdk/compare/185aeb134836c4e464ac1b478e3c1b1ea824e785..c66399cfc006c1aa22303493d99671cbfdcb15fd) until our computer overlord showed me the true way (that is much uglier and harder to read).

PR #661, opened 2024-10-15
Blockwise download error handling

szczys:
Propagate any errors from the user callback out to the initial blockwise download call.

Resolves https://github.com/golioth/firmware-issue-tracker/issues/713

This implements the parts of an existing PR that are still relevant to the current state of the SDK so this PR also:
- Closes #491

### DIFF
--- a/src/coap_blockwise.c
+++ b/src/coap_blockwise.c
@@ -397,7 +397,7 @@ enum golioth_status golioth_blockwise_get(struct golioth_client *client,
     enum golioth_status status = GOLIOTH_ERR_FAIL;
     if (!client || !path || !cb)
     {
-        return status;
+        return GOLIOTH_ERR_NULL;
     }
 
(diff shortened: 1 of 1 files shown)


src/coap_blockwise.c
```diff
@@ -401,9 +395,9 @@ enum golioth_status golioth_blockwise_get(struct golioth_client *client,
                                           void *callback_arg)
 {
     enum golioth_status status = GOLIOTH_ERR_FAIL;
-    if (!client || !path || !cb)
+    if (!client || !path || !path_prefix || !cb)
```
sam-golioth, line 398: In the other CoAP functions (set, get, delete, etc.), we allow `path_prefix` to be `NULL`

sam-golioth (approved):

src/coap_blockwise.c
szczys: Okay, I'll take `path_previx` out of the list (but leave the change for GOLIOTH_ERR_NULL to be returned for other null pointers).

PR #665, opened 2024-10-21
Component Download: implement HIL test for resuming download and verifying component hash

szczys:
- Add HIL test for resuming component download and verifying the hash

I had considered replacing the block download part of the test_block_ops() function, but that uses a non-callback approach to fetching the blocks so I thought it would be best to have both approaches tested.

Resolves https://github.com/golioth/firmware-issue-tracker/issues/44
Resolves https://github.com/golioth/firmware-issue-tracker/issues/716

### DIFF
--- a/tests/hil/tests/ota/test.c
+++ b/tests/hil/tests/ota/test.c
@@ -16,6 +16,7 @@ LOG_TAG_DEFINE(test_ota);
 static golioth_sys_sem_t connected_sem;
 static golioth_sys_sem_t reason_test_sem;
 static golioth_sys_sem_t block_test_sem;
+static golioth_sys_sem_t resume_test_sem;
 
 #define GOLIOTH_OTA_REASON_CNT 10
 #define GOLIOTH_OTA_STATE_CNT 4
@@ -29,6 +30,14 @@ static uint8_t callback_arg = 17;
 static uint8_t block_buf[CONFIG_GOLIOTH_BLOCKWISE_DOWNLOAD_MAX_BLOCK_SIZE];
 
 struct golioth_client *client;
+
+static struct golioth_ota_component stored_component;
+struct resume_ctx
+{
+    bool in_progress;
+    bool fail_pending;
+    golioth_sys_sha256_t sha;
+};
 
 static void log_component_members(const struct golioth_ota_component *component)
 {
@@ -130,7 +139,7 @@ static void test_block_ops(void)
     bool is_last = false;
 
     GLTH_LOGI(TAG,
-              "golioth_ota_size_to_nblocks: %d",
+              "golioth_ota_size_to_nblocks: %zu",
               golioth_ota_size_to_nblocks(
                   (TEST_BLOCK_CNT * CONFIG_GOLIOTH_BLOCKWISE_DOWNLOAD_MAX_BLOCK_SIZE) + 1));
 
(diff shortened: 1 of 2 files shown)


include/golioth/ota.h
```diff
@@ -188,6 +188,25 @@ enum golioth_status golioth_ota_download_component(struct golioth_client *client
                                                    ota_component_block_write_cb cb,
                                                    void *arg);
 
+/// Verify the component sha256 hash supplied by server matches on calculated locally
```
sam-golioth, line 191:
```suggestion
/// Verify the component sha256 hash supplied by server matches one calculated locally
```

include/golioth/ota.h
```diff
@@ -188,6 +188,25 @@ enum golioth_status golioth_ota_download_component(struct golioth_client *client
...
+///
+/// This function compares the component hash received in the OTA manifest to one that is supplied.
+/// Use the golioth_sys_sha256_*() functions (or similar) to calculate the has as blocks are
```
sam-golioth, line 194:
```suggestion
/// Use the golioth_sys_sha256_*() functions (or similar) to calculate the hash as blocks are
```

include/golioth/ota.h
```diff
@@ -188,6 +188,25 @@ enum golioth_status golioth_ota_download_component(struct golioth_client *client
...
+/// hash for the component that was downloaded.
+///
+/// @retval GOLIOTH_OK The \p local_hash matches the component hash recevied from the server.
+/// server.
+/// @retval GOLIOTH_ERR_INVALID_FORMAT The \p local_hash does not match the hash the server
+/// supplied.
+/// @retval GOLIOTH_ERR_NULL One of the supplied parameters was NULL.
+enum golioth_status golioth_ota_component_verify_sha256(
+    const struct golioth_ota_component *component,
+    uint8_t *local_hash);
```
sam-golioth, line 208:
```suggestion
    const uint8_t *local_hash);
```

src/ota.c
```diff
@@ -373,6 +373,29 @@ enum golioth_status golioth_ota_download_component(struct golioth_client *client
...
+{
+    if (!component || !local_hash)
+    {
+        return GOLIOTH_ERR_NULL;
+    }
+
+    unsigned char sha_server[32];
+    golioth_sys_hex2bin(component->hash, strlen(component->hash), sha_server, sizeof(sha_server));
+
+    GLTH_LOG_BUFFER_HEXDUMP(TAG, local_hash, sizeof(local_hash), GOLIOTH_DEBUG_LOG_LEVEL_DEBUG);
```
sam-golioth, line 388: Let's remove the logging here, the application can add logging if they desire.

src/ota.c
```diff
@@ -373,6 +373,29 @@ enum golioth_status golioth_ota_download_component(struct golioth_client *client
...
+enum golioth_status golioth_ota_component_verify_sha256(
+    const struct golioth_ota_component *component,
+    uint8_t *local_hash)
+{
+    if (!component || !local_hash)
+    {
+        return GOLIOTH_ERR_NULL;
+    }
+
+    unsigned char sha_server[32];
```
sam-golioth, line 385: Let's #define a constant for the size of this array.

src/ota.c
```diff
@@ -373,6 +373,29 @@ enum golioth_status golioth_ota_download_component(struct golioth_client *client
                                  &ctx);
 }
 
+enum golioth_status golioth_ota_component_verify_sha256(
+    const struct golioth_ota_component *component,
+    uint8_t *local_hash)
```
sam-golioth, line 378:
```suggestion
    const uint8_t *local_hash)
```

src/ota.c
```diff
@@ -373,6 +373,29 @@ enum golioth_status golioth_ota_download_component(struct golioth_client *client
...
+
+    GLTH_LOG_BUFFER_HEXDUMP(TAG, local_hash, sizeof(local_hash), GOLIOTH_DEBUG_LOG_LEVEL_DEBUG);
+    GLTH_LOG_BUFFER_HEXDUMP(TAG, sha_server, sizeof(sha_server), GOLIOTH_DEBUG_LOG_LEVEL_DEBUG);
+
+    if (memcmp(local_hash, sha_server, sizeof(sha_server)) == 0)
+    {
+        return GOLIOTH_OK;
+    }
+
+    return GOLIOTH_ERR_INVALID_FORMAT;
```
sam-golioth, line 396: This return code doesn't make much sense. The format is valid, the value is wrong. We can use `GOLIOTH_ERR_FAIL` instead.

tests/hil/tests/ota/test.c
```diff
@@ -160,14 +169,118 @@ static void test_block_ops(void)
...
+    }
+
+    struct resume_ctx *ctx = (struct resume_ctx *) arg;
+
+    if (block_idx == 0)
+    {
+        /* This was already done with the create() function
+         * but let's call it again to test the api
+         */
+        golioth_sys_sha256_init(ctx->sha);
```
sam-golioth, line 201: Per my comment on the earlier PR, I don't think it makes sense to have both `create()` and `init()` for the purposes of our port layer.

tests/hil/tests/ota/test.c
```diff
@@ -160,14 +169,118 @@ static void test_block_ops(void)
...
+    }
+
+    if (ctx->fail_pending && block_idx == 2)
+    {
+        ctx->fail_pending = false;
+        return GOLIOTH_ERR_FAIL;
+    }
```
sam-golioth, line 208: Clever!

include/golioth/ota.h
```diff
@@ -188,6 +188,25 @@ enum golioth_status golioth_ota_download_component(struct golioth_client *client
...
+/// @param component The @ref golioth_ota_component pointer from the OTA manifest
+/// @param local_hash An array, exactly 32 bytes in length, containing a locally-calculated sha256
+/// hash for the component that was downloaded.
+///
+/// @retval GOLIOTH_OK The \p local_hash matches the component hash recevied from the server.
+/// server.
+/// @retval GOLIOTH_ERR_INVALID_FORMAT The \p local_hash does not match the hash the server
+/// supplied.
+/// @retval GOLIOTH_ERR_NULL One of the supplied parameters was NULL.
+enum golioth_status golioth_ota_component_verify_sha256(
```
sam-golioth, line 206: Why create this new API instead of just having `golioth_ota_download_component()` calculate and verify the SHA automatically?

szczys: The user may want to verify the hash of the full component after the block download has completed and the data is written somewhere in memory. I would think this is for users who want to roll their own component management outside of the automated download (with resume) that we will eventually have in the OTA service.

Do you prefer it be removed?

sam-golioth: If the user wants to verify the hash, they can extract it from the manifest and do it themselves. This is a simple utility function, I don't think it makes sense to have it as part of our API

szczys: Sounds good to me. I've removed this commit and will rename the PR.

sam-golioth (approved):
Good work @szczys! I had overlooked that we didn't have any test for `golioth_ota_download_component()` yet so this kills two birds with one stone.

PR #666, opened 2024-10-21
kconfig: add OTA symbol that FW_UPDATE depends on

szczys:
Add CONFIG_GOLIOTH_OTA so that the OTA functions may be used without the firmware update functions.

Firmware updates still depend on OTA functions. Platforms that use Kconfig will automatically select OTA. However, for platforms that don't use Kconfig, ota.c will build when either CONFIG_GOLIOTH_OTA or CONFIG_GOLIOTH_FW_UPDATE are selected.

This is not a panacea but there is some savings to be had if an application wants to work with OTA componentes without fw_update:
- 60 bytes of flash saved
- 640 bytes of RAM saved

Resolves https://github.com/golioth/firmware-issue-tracker/issues/709


src/Kconfig
```diff
@@ -4,12 +4,13 @@
...
+    bool "Golioth Over-the-Air Update Service"
     help
-        Enable the Golioth Firmware Update service
+        Enable the Golioth OTA service to manage the transfer of OTA components between server and
+        device. See also: GOLIOTH_FW_UPDATE
 
-if GOLIOTH_FW_UPDATE
+if GOLIOTH_OTA
 
 config GOLIOTH_OTA_THREAD_STACK_SIZE
```
sam-golioth, line 15: This is actually used by the fw_update module, not OTA. We should have it guarded by `GOLIOTH_FW_UPDATE` instead and also rename it.

src/Kconfig
```diff
@@ -48,7 +49,14 @@ config GOLIOTH_OTA_OBSERVATION_RETRY_MAX_DELAY_S
         backoff doubling between retry attempts. The delay value will increase until this limit is
```
sam-golioth, line 49: Can't comment on the exact line, but `GOLIOTH_OTA_OBSERVATION_RETRY_MAX_DELAY_S` is also used by the firmware update module, not OTA.

szczys:
Apologies, I didn't look to see where those encapsulated symbols were used but I definitely should have. This is sorted out now.

src/Kconfig
```diff
@@ -37,7 +32,16 @@ config GOLIOTH_OTA_MAX_NUM_COMPONENTS
...
+endif # GOLIOTH_OTA
+
+config GOLIOTH_FW_UPDATE
+    bool "Golioth Firmware Update service"
+    select GOLIOTH_OTA
+    help
+        Enable the Golioth Firmware Update service to automatically manage firmware updates
+        utilizing functionality of the Golioth OTA service.
+
+config GOLIOTH_FW_UPDATE_OBSERVATION_RETRY_MAX_DELAY_S
```
sam-golioth, line 44: Should this also be guarded by `if GOLIOTH_FW_UPDATE`?

sam-golioth (approved):

src/Kconfig
szczys: Yes, thanks for catching!

PR #669, opened 2024-11-12
Run callbacks on request cancellation

szczys:
- Run callbacks when cancelling requests
- Add `GOLIOTH_ERR_COAP_RESPONSE_CODE` to indicate a non-success CoAP response code was received
- Pass SDK errors to async callbacks using the response->status member

Resolves https://github.com/golioth/firmware-issue-tracker/issues/727
Resolves https://github.com/golioth/firmware-issue-tracker/issues/728


CHANGELOG.md
```diff
@@ -15,6 +15,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   include `block_buffer_len` for the actual length of data and
   `negotiated_block_size` to indicate the maximum block size (may be
   used along with 'block_idx' to calculate a byte offset).
+- Asynchronous callbacks now return error information for both the
+  Golioth Firmware SDK and the CoAP response code. When response->status
+  is GOLIOTH_OK or GOLIOTH_ERR_COAP_REPONSE_CODE, read the
+  response->status_class and response->status_code for the CoAP response
+  code. All other response->status codes indicate an SDK error and the
+  CoAP response codes are invalid/undefined.
```
szczys:
```suggestion
  code. All other response->status codes indicate an error unrelated
  to the CoAP response codes which are invalid/undefined in this case.
```

include/golioth/client.h
```diff
@@ -146,6 +146,11 @@ typedef void (*golioth_client_event_cb_fn)(struct golioth_client *client,
 /// The callback function should check response->status to determine which case it is (response
 /// received or timeout).
 ///
+/// When response->status is GOLIOTH_OK or GOLIOTH_ERR_COAP_RESPONSE_CODE, the
+/// response->status_class and response->status_code may be read to determine the CoAP response code
+/// received. All other response->status codes indicate an SDK error code and the CoAP status
+/// members are invalid/undefined.
```
szczys, line 153:
```suggestion
/// received. All other response->status codes indicate an SDK error status and the CoAP response code
/// members are invalid/undefined.
```

include/golioth/client.h
```diff
@@ -165,6 +170,11 @@ typedef void (*golioth_get_cb_fn)(struct golioth_client *client,
 /// response never received). The callback function should check response->status to determine which
 /// case it is (response received or timeout).
 ///
+/// When response->status is GOLIOTH_OK or GOLIOTH_ERR_COAP_RESPONSE_CODE, the
+/// response->status_class and response->status_code may be read to determine the CoAP response code
+/// received. All other response->status codes indicate an SDK error code and the CoAP status
+/// members are invalid/undefined.
```
szczys, line 177:
```suggestion
/// received. All other response->status codes indicate an SDK error status and the CoAP response code
/// members are invalid/undefined.
```

include/golioth/client.h
```diff
@@ -188,6 +198,11 @@ typedef void (*golioth_get_block_cb_fn)(struct golioth_client *client,
 /// received or timeout). If response->status is GOLIOTH_OK, then the set or delete request
 /// was successful.
 ///
+/// When response->status is GOLIOTH_OK or GOLIOTH_ERR_COAP_RESPONSE_CODE, the
+/// response->status_class and response->status_code may be read to determine the CoAP response code
+/// received. All other response->status codes indicate an SDK error code and the CoAP status
+/// members are invalid/undefined.
```
szczys, line 205:
```suggestion
/// received. All other response->status codes indicate an SDK error status and the CoAP response code
/// members are invalid/undefined.
```

src/coap_client_zephyr.c
```diff
@@ -330,10 +314,15 @@ static int golioth_coap_cb(struct golioth_req_rsp *rsp)
...
     }
 
+    if (req->type == GOLIOTH_COAP_REQUEST_OBSERVE)
+    {
+        // don't free observations so we can reestablish later
+        return rsp->codes.status;
+    }
+
 free_req:
     free(req);
```
szczys, line 324:
```suggestion
    if (req->type == GOLIOTH_COAP_REQUEST_OBSERVE)
    {
        free(req);
    }
```

Gate this `free()` call instead of returning above. This is safer and less likely to be accidentally missed in future changes.

src/coap_client_zephyr.c
```diff
@@ -330,10 +314,15 @@ static int golioth_coap_cb(struct golioth_req_rsp *rsp)
         golioth_sys_sem_destroy(req->request_complete_ack_sem);
     }
 
+    if (req->type == GOLIOTH_COAP_REQUEST_OBSERVE)
+    {
+        // don't free observations so we can reestablish later
+        return rsp->codes.status;
+    }
+
```
szczys, line 322:
```suggestion
```

Instead of checking here, gate the `free()` call below.

src/coap_blockwise.c
```diff
@@ -245,21 +245,13 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
...
-        bool local_error =
-            (status != GOLIOTH_OK && (ctx->block_idx == 0 || ctx->response.status == GOLIOTH_OK));
-
-        if (!local_error)
+        /* Pass the local error condition to the callback if an error is not already set */
+        if (status != GOLIOTH_OK && ctx->response.status == GOLIOTH_OK)
         {
-            set_cb(client, &ctx->response, path, callback_arg);
+            ctx->response.status = status;
         }
```
sam-golioth, line 252: This feels a little confusing and hacky to me. Can you explain more the cases where the return value would be non-OK but the response status would be OK?

Or more generally what we're trying to accomplish here?

include/golioth/client.h
```diff
@@ -49,14 +49,15 @@ enum golioth_content_type
...
     /// Status to indicate whether a response was received
     ///
     /// One of:
+    ///      GOLIOTH_OK (CoAP codes was in the success class: 2.XX)
+    ///      GOLITOH_ERR_COAP_RESPONSE (CoAP error; anything other than 2.xx. Check CoAP codes)
     ///      GOLIOTH_ERR_TIMEOUT (no response received from server)
-    ///      GOLIOTH_OK (2.XX)
-    ///      GOLIOTH_ERR_FAIL (anything other than 2.XX)
+    ///      GOLIOTH_ERR_FAIL (All other failures not related to a response)
     enum golioth_status status;
```
sam-golioth, line 61: I think we could make things even clearer by removing the golioth_status from this struct and passing it separately, so that this struct only contains CoAP response codes.

src/coap_blockwise.c
sam-golioth: Yeah, that's a little better. But I think part of the confusion is actually related to keeping the golioth status inside the response struct. Because this code feels like it's changing the server response.

To make it even more clear when the CoAP response holds something valid, I think we can do something like 

```
struct golioth_response *rsp_ptr = NULL;
if (status == GOLIOTH_OK || status  == GOLIOTH_ERR_COAP_RESPONSE)
{
    rsp_ptr = &ctx->response;
}
set_cb(client, status, rsp_ptr, path, callback_arg);
```

include/golioth/client.h
sam-golioth: Most of the confusion actually arises inside the SDK. The problem is that they're _not_ always tightly coupled. When the `status` field holds a value that is neither `GOLIOTH_OK` nor `GOLIOTH_ERR_COAP_RESPONSE`, they're not coupled at all.

examples/common/golioth_basics.c
```diff
@@ -85,16 +85,17 @@ static enum golioth_rpc_status on_multiply(zcbor_state_t *request_params_array,
```
szczys, line 85: This commit message needs to be updated for the new member names: `status_class` used instead of `class` (which is a protected name in CPP).

This commit should also have a breaking change entry in the CHANGELOG.

include/golioth/client.h
```diff
@@ -142,38 +136,59 @@ typedef void (*golioth_client_event_cb_fn)(struct golioth_client *client,
...
-/// The callback function should check response->status to determine which case it is (response
-/// received or timeout).
+/// Will be called when a response is received, on timeout (i.e. response never received), or when
+/// the request is cancelled (e.g. when the Golioth client is stopped). The callback function should
+/// check \p status to determine which case triggered the callback.
+///
+/// The CoAP response code is available by reading class (2.xx) and detail (x.00) from \p
+/// coap_rsp_code:
+/// - When \p status is GOLIOTH_OK, \p coap_rsp_code->class will be a 2.XX success code.
+/// - When \p status is GOLIOTH_COAP_RESPONSE_CODE, check \p coap_rsp_code->class for a non-success
```
szczys, line 146:
```suggestion
/// The CoAP response code is available by reading status_class (2.xx) and status_detail (x.00) from \p
/// coap_rsp_code:
/// - When \p status is GOLIOTH_OK, \p coap_rsp_code->status_class will be a 2.XX success code.
/// - When \p status is GOLIOTH_COAP_RESPONSE_CODE, check \p coap_rsp_code->status_class for a non-success
```

include/golioth/client.h
```diff
@@ -142,38 +136,59 @@ typedef void (*golioth_client_event_cb_fn)(struct golioth_client *client,
...
-/// case it is (response received or timeout).
+/// Will be called repeatedly, once for each block received from the server, on timeout (i.e.
+/// response never received), or when the request is cancelled (e.g. when the Golioth client is
+/// stopped). The callback function should check \p status to determine which case triggered the
+/// callback.
+///
+/// The CoAP response code is available by reading class (2.xx) and detail (x.00) from \p
+/// coap_rsp_code:
+/// - When \p status is GOLIOTH_OK, \p coap_rsp_code->class will be a 2.XX success code.
+/// - When \p status is GOLIOTH_COAP_RESPONSE_CODE, check \p coap_rsp_code->class for a non-success
```
szczys, line 176: Same changes as above.

include/golioth/client.h
```diff
@@ -183,17 +198,26 @@ typedef void (*golioth_get_block_cb_fn)(struct golioth_client *client,
...
-/// received or timeout). If response->status is GOLIOTH_OK, then the set or delete request
-/// was successful.
+/// Will be called when a response is received, on timeout (i.e. response never received), or when
+/// the request is cancelled (e.g. when the Golioth client is stopped). The callback function should
+/// check \p status to determine which case triggered the callback.
+///
+/// The CoAP response code is available by reading class (2.xx) and detail (x.00) from \p
+/// coap_rsp_code:
+/// - When \p status is GOLIOTH_OK, \p coap_rsp_code->class will be a 2.XX success code.
+/// - When \p status is GOLIOTH_COAP_RESPONSE_CODE, check \p coap_rsp_code->class for a non-success
```
szczys, line 208: Same changes as above.

src/coap_blockwise.c
```diff
@@ -339,7 +344,9 @@ static void on_block_rcvd(struct golioth_client *client,
     ctx->is_last = is_last;
     memcpy(ctx->buffer, payload, payload_size);
     ctx->rcvd_bytes = payload_size;
-    ctx->status = response->status;
+    ctx->status = status;
+    ctx->coap_rsp_code.status_class = coap_rsp_code->status_class;
+    ctx->coap_rsp_code.status_detail = coap_rsp_code->status_detail;
```
szczys, line 349: Do we ever use this in the blockwise get operation?

src/coap_client_libcoap.c
```diff
@@ -50,7 +52,8 @@ static void notify_observers(const coap_pdu_t *received,
             && (0 == memcmp(rcvd_token.s, obs_info->req.token, obs_info->req.token_len)))
         {
             callback(client,
-                     response,
+                     status,
+                     coap_rsp_code,
```
szczys, line 56: Does this need to be tested fro returning NULL based on the status value?

examples/common/golioth_basics.c
sam-golioth: We should use `code_class` and `code_detail` to align with the CoAP spec:

```
Code:  8-bit unsigned integer, split into a 3-bit class (most
      significant bits) and a 5-bit detail (least significant bits),
      documented as "c.dd" where "c" is a digit from 0 to 7 for the
      3-bit subfield and "dd" are two digits from 00 to 31 for the 5-bit
      subfield.  The class can indicate a request (0), a success
      response (2), a client error response (4), or a server error
      response (5).  (All other class values are reserved.)  As a
      special case, Code 0.00 indicates an Empty message.  In case of a
      request, the Code field indicates the Request Method; in case of a
      response, a Response Code.  Possible values are maintained in the
      CoAP Code Registries (Section 12.1).  The semantics of requests
      and responses are defined in Section 5.
```

`status` has no canonical meaning in this context.

src/coap_blockwise.c
szczys: We don't use this for a blockwise get so they do not need to be set in this callback.

sam-golioth (approved):
Great work @szczys !

src/coap_blockwise.c
szczys: Removing line 347 caused block download to immediately fail for espidf. The status is passed back from the context to the falling function and still needs to be assigned here. However, the coap code is not used and doesn't need to be assigned.

PR #672, opened 2024-11-14
coap_client_zephyr: ensure keepalive timer reset on observation

sam-golioth:
Ensure that we reset the keepalive timer when we receive an observation response.

### DIFF
--- a/src/coap_client_zephyr.c
+++ b/src/coap_client_zephyr.c
@@ -306,8 +306,9 @@ static int golioth_coap_cb(struct golioth_req_rsp *rsp)
                 req->observe
                     .callback(client, &response, req->path, rsp->data, rsp->len, req->observe.arg);
             }
-            /* There is no synchronous version of observe request */
-            return 0;
+            /* There is no synchronous version of observe request, req->request_complete_event will
+               always be NULL. */
+            break;
     }
 
     if (CONFIG_GOLIOTH_COAP_KEEPALIVE_INTERVAL_S > 0)


src/coap_client_zephyr.c
```diff
@@ -306,8 +306,9 @@ static int golioth_coap_cb(struct golioth_req_rsp *rsp)
                 req->observe
                     .callback(client, &response, req->path, rsp->data, rsp->len, req->observe.arg);
             }
-            /* There is no synchronous version of observe request */
-            return 0;
+            /* There is no synchronous version of observe request, req->request_complete_event will
+               always be NULL. */
+            break;
```
szczys, line 311: If we break here, the request memory will be freed at the bottom of the function and will cause a use after free error when trying to reestablish observations after a client start/stop. I tested this with a mimxrt1024_evk on the RPC integration test and received an MPU fault.

However, I made the same change (not for the same reason) in the callbacks on cancellation PR: https://github.com/golioth/golioth-firmware-sdk/pull/669/commits/1f7fbfb2d9bf3f78d624e1f0654e5e3f3f228fdf

The commit before that change adds a conditional check just before the free operation that returns if the type is an observation:
https://github.com/golioth/golioth-firmware-sdk/pull/669/commits/f671be0b82eab926ffa0bcf2e90b1d45b08460eb#diff-1478f24ca0f9d927604108f856f882c69c95561f408c7c9a26ac1d823b8db085R314

szczys (changes requested):

szczys:
Should we also reset the keepalive for EMPTY and OBSERVE_RELEASE? 

https://github.com/golioth/golioth-firmware-sdk/blob/5ece8c9dad89006dbb94e529c27d10e805ed6df4/src/coap_client_zephyr.c#L262

src/coap_client_zephyr.c
sam-golioth: Great catch! I agree your changes address this better, will close this PR.

PR #674, opened 2024-11-15
hil: ota: use 1 Hz polling for reason and state test

szczys:
Use 1 Hz polling in pytest to catch reason and state changes more efficiently.

This removes an arbitrary delay, instead trying multiple times to get the device's reported reason and state codes. On the final timeout, the asserts will always be run so we retain helpful logging messages. This PR also reduces the device delay by 3 seconds, which equates to 30 seconds saved across all reason codes.

### DIFF
--- a/tests/hil/tests/ota/test.c
+++ b/tests/hil/tests/ota/test.c
@@ -98,7 +98,7 @@ static void test_reason_and_state(void)
         {
             GLTH_LOGI(TAG, "OTA status reported successfully");
         }
-        golioth_sys_msleep(7000);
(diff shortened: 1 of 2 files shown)


sam-golioth:
Changes look good! Can we run the test a few times to see how stable it is?

szczys:
> Changes look good! Can we run the test a few times to see how stable it is?

I ran the entire suite twice (once with original PR and once last night on a force-push) and it passed both times. I ran the nsim-integration and linux-integration just now and those passed as well.

sam-golioth (approved):

PR #679, opened 2024-11-19
certs: replace ISRG Root X2 with Golioth Root X1

hasheddan:
Replaces the ISRG Root X2 root CA certificate with the Golioth Root X1 CA certificate. ISRG Root X1 is currently in use by Golioth servers, and ISRG Root X2 had been previously added to the firmware SDK to smooth the transition to using an elliptic curve chain in the future. However, Golioth has decided to move to managing its own elliptic curve chain and will move to using Golioth Root X1 as the root in the future.


hasheddan:
Replaced by https://github.com/golioth/golioth-firmware-sdk/pull/680

PR #681, opened 2024-11-20
examples: fw_update: pytest: drop --west-board and ${west_board} use

mniestroj:
Don't use (and don't require) `--west-board` CLI argument and/or
`${west_board}` environment variable. Use 'platform' provided by twister
harness instead. This simplifies twister setup.


sam-golioth (approved):
I love a PR that's mostly deletions!

PR #689, opened 2024-11-22
test: fw_update: use first platform specified

hasheddan:
Platform is a list argument and can be specified multiple times. We currently only ever invoke pytest with a single platform and it should be used to fetch the blueprint ID for the fw_update sample test.

### DIFF
--- a/examples/zephyr/fw_update/pytest/conftest.py
+++ b/examples/zephyr/fw_update/pytest/conftest.py
@@ -2,13 +2,14 @@ import golioth
 import os
 import pytest
 
-UPDATE_VERSION = '255.8.9'
-UPDATE_PACKAGE = 'main'
+UPDATE_VERSION = "255.8.9"
+UPDATE_PACKAGE = "main"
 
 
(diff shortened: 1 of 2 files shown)


hasheddan (commented):
ref: https://github.com/zephyrproject-rtos/zephyr/blob/da713a2b86c00b1d3167e35a75debf16db5d0931/scripts/pylib/twister/twisterlib/environment.py#L615

examples/zephyr/fw_update/pytest/conftest.py
```diff
@@ -17,13 +18,15 @@ async def fw_info():
 
 @pytest.fixture(scope="module")
 async def blueprint_id(project, request):
-    bp_name = request.config.option.platform.replace('/','_')
+    bp_name = request.config.option.platform[0].replace("/", "_")
```
hasheddan, line 21: This is actually a different `--platform` than the one we pass to twister. The `--platform` here is from the value that twister passes to pytest.

PR #693, opened 2024-11-27
fw_update: add resume download functionality

szczys:
When an error occurs during a firmware block download, the download will now resume from that block instead of starting from the beginning.

Each block is included in a sha256 hash calculation during the download and compared to the server-supplied hash to verify a successful download. This comparison is made while the data is in RAM, before it is written to flash.

If a flash write error occurs during block download, the firmware download is considered a failure and will not try to resume from that block.

Resolves: https://github.com/golioth/firmware-issue-tracker/issues/688
Resolves: https://github.com/golioth/firmware-issue-tracker/issues/698

### DIFF
--- a/src/fw_update.c
+++ b/src/fw_update.c
@@ -31,6 +31,8 @@ struct download_progress_context
     golioth_sys_sha256_t sha;
 };
 
+#define FW_UPDATE_RESUME_DELAY_S 15
+
 static enum golioth_status fw_write_block_cb(const struct golioth_ota_component *component,
                                              uint32_t block_idx,
                                              uint8_t *block_buffer,
@@ -52,12 +54,15 @@ static enum golioth_status fw_write_block_cb(const struct golioth_ota_component 
                                                         negotiated_block_size * block_idx,
                                                         component->size);
 
-    ctx->bytes_downloaded += block_buffer_len;
-    golioth_sys_sha256_update(ctx->sha, block_buffer, block_buffer_len);
-
-    if (is_last)
+    if (status == GOLIOTH_OK)
     {
-        fw_update_post_download();
+        ctx->bytes_downloaded += block_buffer_len;
+        golioth_sys_sha256_update(ctx->sha, block_buffer, block_buffer_len);
+
+        if (is_last)
+        {
+            fw_update_post_download();
+        }
     }
 
     return status;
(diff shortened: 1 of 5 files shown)


src/fw_update.c
```diff
@@ -180,6 +185,47 @@ static void fw_report_and_observe(void)
     }
 }
 
+static enum golioth_status fw_verify_component_hash(
+    struct download_progress_context *ctx,
+    const uint8_t server_sha256[GOLIOTH_OTA_COMPONENT_BIN_HASH_LEN])
+{
+    uint8_t calc_sha256[32];
```
szczys, line 192:
```suggestion
    uint8_t calc_sha256[GOLIOTH_OTA_COMPONENT_BIN_HASH_LEN];
```

src/fw_update.c
```diff
@@ -52,6 +52,11 @@ static enum golioth_status fw_write_block_cb(const struct golioth_ota_component
                                                         negotiated_block_size * block_idx,
                                                         component->size);
 
+    if (status != GOLIOTH_OK)
+    {
+        return status;
+    }
+
     ctx->bytes_downloaded += block_buffer_len;
```
sam-golioth: We can make this clearer by making sure there's only one return statement in this function. Either changing this `return` into ` goto`, or removing the added `if` and instead wrapping the rest  of the function in `if (status == GOLIOTH_OK)`:

```
if (status == GOLIOTH_OK)
{
    ctx->bytes_downloaded += block_buffer_len;

    if (is_last)
    {
        fw_update_post_download();
    }
}

return status;
```

src/fw_update.c
```diff
@@ -25,6 +25,8 @@ static golioth_fw_update_state_change_callback _state_callback;
 static void *_state_callback_arg;
 static struct golioth_fw_update_config _config;
 
+#define FW_MAX_BLOCK_RESUME_BEFORE_FAIL 15
```
sam-golioth, line 28: How did you arrive at this number?

src/fw_update.c
```diff
@@ -243,32 +300,90 @@ static void fw_update_thread(void *arg)
...
         uint64_t start_time_ms = golioth_sys_now_ms();
-        size_t bytes_downloaded = 0;
+        download_ctx.bytes_downloaded = 0;
+        uint32_t next_block = 0;
+
+        if (!download_ctx.sha)
+        {
+            golioth_sys_sha256_destroy(download_ctx.sha);
+        }
+        download_ctx.sha = golioth_sys_sha256_create();
```
sam-golioth, line 310: Are we destroying this sha anywhere other than above? We should destroy it anytime we complete or fail a firmware update.

src/fw_update.c
```diff
@@ -264,11 +298,18 @@ static void fw_update_thread(void *arg)
                                                 _main_component->version,
                                                 GOLIOTH_SYS_WAIT_FOREVER);
 
+            golioth_sys_sha256_destroy(download_ctx.sha);
```
sam-golioth, line 380: Can we call `golioth_sys_sha256_destroy()` on line 278 after the call to `golioth_sys_sha256_finish()` so that we only need one call?

src/fw_update.c
```diff
@@ -257,14 +301,25 @@ static void fw_update_thread(void *arg)
...
+            {
+                GLTH_LOGE(TAG, "Failed to store OTA component");
+                break;
+            }
+
+            if (err == GOLIOTH_OK)
+            {
+                break;
+            }
+            else
```
sam-golioth, line 352: Should we include a delay here?

src/fw_update.c
sam-golioth: I see, in a later commit the call to `golioth_sys_sha256_finish()` is moved to another function 👍

src/fw_update.c
szczys: We could. Is there a reason to include a delay?

At the top of the loop we will add a new request to the mailbox so this will get put at the end of the line if there are any other network operations in progress. If the client is not running this will just queue the operation until the client starts again.

src/fw_update.c
szczys: I have some consternation about this approach because a few lines down there is a call to `fw_update_validate()` and I think this should probably replaced that. I'm trying to limit the scope of this PR so figured we could clean that up in future work.

sam-golioth: Do you mean that comparing the SHA256 should replace `fw_update_validate()`? I think it's useful to have a point where applications can insert their own custom validation logic.

For my original comment, we can leave it as is given that the organization of the code changed in the later commit.

src/fw_update.c
sam-golioth: If there are error conditions that cause the request to fail immediately (like the mailbox being full), this will be a tight loop and quickly exhaust the retries. Even if there is nothing else going on, whatever the transient reason that caused the firmware update to fail could exist for some period of time. I think it's useful here to have a small delay that gives whatever error condition time to pass. What do you think?

src/fw_update.c
szczys: Sorry I wasn't clear. I don't want to replace the call to `fw_update_validate()` but I think it should be grouped with the sha256 verfication. I started to do so with this PR but it becomes a bigger change than I had anticipated. I'll open an issue to track this and we can clean it up (hopefully) as part of the OTA stability epic.

src/fw_update.c
szczys: I see, that's a good point.

I added a delay of 15 seconds. When combined with 15 max resumes on a single block we have a delay of up to 3:45 which should be ample for a cellular connection to be re-established before the download is considered a failure.

sam-golioth: I'm starting to feel like I can guess your favorite number 😉

src/fw_update.c
sam-golioth: Sounds good!

sam-golioth (approved):

PR #695, opened 2024-12-03
fw_update: do not drop manifests based on state; store target components

szczys:
Allow a new manifest to be received regardless of the state of the fw_update thread. This removes the need to drop a new manifest during the downloading step. To make this possible, a new version of the target component is copied from the manifest for the fw_update thread to use during the download process.

Resolves: https://github.com/golioth/firmware-issue-tracker/issues/748

### DIFF
--- a/src/fw_update.c
+++ b/src/fw_update.c
@@ -17,19 +17,24 @@
 
 LOG_TAG_DEFINE(golioth_fw_update);
 
-static struct golioth_client *_client;
-static golioth_sys_sem_t _manifest_rcvd;
-static struct golioth_ota_manifest _ota_manifest;
-static const struct golioth_ota_component *_main_component;
-static golioth_fw_update_state_change_callback _state_callback;
-static void *_state_callback_arg;
-static struct golioth_fw_update_config _config;
+struct fw_update_component_context
+{
+    struct golioth_fw_update_config config;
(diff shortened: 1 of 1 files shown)


src/fw_update.c
```diff
@@ -17,7 +17,8 @@
 
 LOG_TAG_DEFINE(golioth_fw_update);
 
-struct fw_update_component_context {
+struct fw_update_component_context
+{
```
szczys, line 21: Move this linter fix to the commit in which it was introduced

src/fw_update.c
```diff
@@ -476,7 +476,7 @@ void golioth_fw_update_init_with_config(struct golioth_client *client,
     _client = client;
     _component_ctx.config = *config;
     _manifest_update_mut = golioth_sys_mutex_create();  // never destroyed
-    _manifest_rcvd = golioth_sys_sem_create(1, 0);  // never destroyed
+    _manifest_rcvd = golioth_sys_sem_create(1, 0);      // never destroyed
```
szczys, line 479: Move this linter fix to the commit in which it was introduced

src/fw_update.c
sam-golioth, line 1: Weird typo in the signoff message in the commit: "Mie Szczys"

sam-golioth (approved):

PR #696, opened 2024-12-04
ota: add reason codes related to fw_update

szczys:
We need more reason codes to better indicate behavior added to the SDK. This PR seeks alignment from the firmware and cloud teams as the codes will need to be added to both the SDK and the Console.

Please view the comments for each new code in the commit. I think these should align with what is shown in the console so adding your thoughts on those lines is appreciated!

## Here is a rundown of how I think these codes will be used

1. Firmware download begins, current behavior remains the same:
    ```
    GOLIOTH_OTA_STATE_DOWNLOADING : GOLIOTH_OTA_REASON_READY
    ```

2. Failure while downloading a block (client restart or request timeout):
    ```
    GOLIOTH_OTA_STATE_DOWNLOADING : GOLIOTH_OTA_REASON_BLOCK_FAILED
    ```

3. Previous failure does not cause the download to fail... we will immediately try to resume. This combined with step two means that repeated block failures followed by resumes will not be hidden by deduplication on the cloud.
    ```
    GOLIOTH_OTA_STATE_DOWNLOADING : GOLIOTH_OTA_REASON_RESUMING
    ```

4. Block is downloaded but cannot be written storage (e.g. flash memory)
    ```
    GOLIOTH_OTA_STATE_DOWNLOADING : GOLIOTH_OTA_REASON_IO
    ```

5. Previous error is not resumable, block download fails immediately but will enter a state of retry from the beginning with backoff delay. This is not yet implemented but the plan is that in this state a new manifest can still be received, cancelling the retry.
    ```
    GOLIOTH_OTA_STATE_IDLE : GOLIOTH_OTA_REASON_AWAIT_RETRY
    ```

### DIFF
--- a/include/golioth/ota.h
+++ b/include/golioth/ota.h
@@ -68,6 +68,10 @@ enum golioth_ota_reason
     GOLIOTH_OTA_REASON_FIRMWARE_UPDATE_FAILED,
     /// Protocol not supported
     GOLIOTH_OTA_REASON_UNSUPPORTED_PROTOCOL,
+    /// IO error while trying to store component
+    GOLIOTH_OTA_REASON_IO,
+    /// Awaiting retry
+    GOLIOTH_OTA_REASON_AWAIT_RETRY,
 };
 
 /// A component/artifact within an OTA manifest


include/golioth/ota.h
```diff
@@ -68,6 +68,14 @@ enum golioth_ota_reason
     GOLIOTH_OTA_REASON_FIRMWARE_UPDATE_FAILED,
     /// Protocol not supported
     GOLIOTH_OTA_REASON_UNSUPPORTED_PROTOCOL,
+    /// IO error while trying to store component
+    GOLIOTH_OTA_REASON_IO,
```
hasheddan, line 72: @szczys we also have the `NOT_ENOUGH_FLASH_MEMORY` and `OUT_OF_RAM` reasons. It looks like there is [one use of the former](https://github.com/golioth/golioth-firmware-sdk/blob/5434b10fda4321b889478cde2da0fc9f5059a432/src/fw_update.c#L369) and none of the latter. Do you anticipate the being used in different scenarios or eventually replacing those?

include/golioth/ota.h
```diff
@@ -68,6 +68,14 @@ enum golioth_ota_reason
...
+    /// Failure to download block
+    GOLIOTH_OTA_REASON_BLOCK_FAILED,
+    /// Resuming the interrupted download
+    GOLIOTH_OTA_REASON_RESUMING,
```
hasheddan, line 76: I'm curious if changing reason in this case is worth it. For example, if we are downloading 100 blocks and every 5th block download fails, it might create more noise than is helpful. I think the key thing we want to communicate is:
1. If we are downloading, are we progressing?
2. If we are not progressing, why?

If a block fails and it is retry-able, then I think I would still consider that progressing, until we eventually got into backoff due to repeated failures (then we would move to `AWAIT_RETRY`). If we want to communicate how often that a block is getting re-requested (i.e. network stats), that is something we could observe and report on server-side without needing to have the device report it. Do you think that would accomplish your goals here?

include/golioth/ota.h
szczys: I discussed this with @sam-golioth and he suggested adding the new code without removing the others as these are available for users to use in their own implementations.

include/golioth/ota.h
szczys: I think it would accomplish the same goal. There is something to be said for having the actual report from the device to know its state. I think you're asking due to how the history will display on the console?

We're trying to communicate how often the block download is resuming as not doing so would mask spotty connection and lead to questions about data usage. We need to send alternating messages to avoid deduplication on the server from masking repeated messages.

hasheddan: > I think you're asking due to how the history will display on the console?

I'm actually less concerned about presentation -- presentation can be adapted to most effectively present useful information based on the data that we have. I'm more concerned about reporting more data on an already spotty connection. If the goal is communicating spotty connection statistics then I think we could do as good or better using server-side metrics and / or supporting reporting statistics from a device after an OTA download completes.

include/golioth/ota.h
hasheddan: Just to confirm, the reason we would introduce this less specific reason is that it allows us to report when we aren't sure exactly where data is being written?

sam-golioth: The use of `GOLIOTH_OTA_REASON_NOT_ENOUGH_FLASH_MEMORY` was added recently, before that we weren't specifying any reason other than `GOLIOTH_OTA_REASON_FIRMWARE_UPDATE_FAILED`. In it's current usage, we're translating from `GOLIOTH_ERR_IO` (an internal error code) to `GOLIOTH_OTA_REASON_NOT_ENOUGH_FLASH_MEMORY` which makes assumptions about the implementation of the port layer that we probably shouldn't. We should either continue to report `GOLIOTH_OTA_REASON_FIRMWARE_UPDATE_FAILED` or use the new `GOLIOTH_OTA_REASON_IO` code.

Part of the confusion here is that these error codes (improperly, imo) mix SDK errors with errors returned from user callbacks (or the port layer in the case where the fw_update module is the user of the OTA service, but the port layer can also be customized and overridden by customers). The SDK errors are under our control, we know the failure modes, and we can log them precisely. The error codes that indicate issues in user callbacks (specifically `OUT_OF_FLASH` and `OUT_OF_RAM`) are speculative - we're trying to predict what errors a user supplied component may return, and then use these error codes to provide information to the user to help them debug their own code. Fundamentally I think that's a losing game, and we should instead be using the OTA reason codes to indicate that an error occurred in the user callback while providing a flexible mechanism that allows users to provide more information about that error in a way that is useful for their code. In my opinion, that flexible mechanism is either going to be the Golioth logging service or a third-party observability platform through pipelines. Limiting users to pre-defined error codes is always going to cause some friction and confusion trying to get square pegs through round holes.

The main reason for keeping existing error codes around is to not mess up the enum numbering and break compatibility with previous versions of the SDK. We could rename the enums in the SDK to something like `GOLIOTH_OTA_REASON_DEPRECATED`, but as long as we have to keep the values around anyway we might as well make them available for use.

hasheddan: > Fundamentally I think that's a losing game, and we should instead be using the OTA reason codes to indicate that an error occurred in the user callback while providing a flexible mechanism that allows users to provide more information about that error in a way that is useful for their code. In my opinion, that flexible mechanism is either going to be the Golioth logging service or a third-party observability platform through pipelines. Limiting users to pre-defined error codes is always going to cause some friction and confusion trying to get square pegs through round holes.

This is the basis of my concerns around expansion of states / reasons here. I think we should either have a very narrow state machine where we are effectively saying "use this if you want to take advantage of some of the automatic insights Golioth's OTA service offers", otherwise emit data to one of the more flexible services you mention. I'd almost go so far to say that the existing reasons we have should be managed through a more flexible service. They feel more like a second level observability indicator ("why is it happening") rather than a primary ("what is happening"). It is also a little strange that any permutation of state and reason can be submitted, when clearly some of the reasons are not applicable to all states.

As far as the move to using the `GOLIOTH_OTA_REASON_IO`, that makes sense to me and I appreciate the attention to backwards compatibility. I do think that this list could continue to grow without a clear definition of the level of specificity we are looking for though, so I think it makes sense to explore whether we can find a better balance of combining a rigid happy path with flexibility that allows for users to determine their own required level of specificity.

szczys:
After much discussion with @hasheddan  and @sam-golioth we've arrived at the following:

Add two new reason codes:
```
GOLIOTH_OTA_REASON_IO
GOLIOTH_OTA_REASON_AWAIT_RETRY
```

## Usage examples:

1. Only report the first time a single block fails, don't report any other block failures or retries until this component download completes or fails (sending it to retry with backoff):

```
GOLIOTH_OTA_STATE_DOWNLOADING : GOLIOTH_OTA_REASON_CONNECTION_LOST
```

2. Only report the first time a block is resumed, not again for this or any other block

``` 
GOLIOTH_OTA_STATE_DOWNLOADING : GOLIOTH_OTA_REASON_READY
```

3. After device fails, report when it is awaiting a resume (due to backoff):

```
GOLIOTH_OTA_STATE_IDLE: GOLIOTH_OTA_REASON_AWAIT_RETRY
```

We envision that repeated block resumes can be detected and signaled from the cloud side, removing the need to repeatedly send status/reason updates from the device as long as the download continues to progress.

hasheddan (approved):

sam-golioth (approved):

mniestroj (approved):

PR #697, opened 2024-12-05
coap: tokens: store as uint64_t instead of as array

szczys:
Store CoAP tokens as uint64_t values instead of as 8-byte arrays.


szczys:
This approach feels rather awkward. For now we're going to keep the token as an array.

Closes https://github.com/golioth/firmware-issue-tracker/issues/750

PR #700, opened 2024-12-09
Use _ret variants of mbedtls functions introduced in 2.7.0

atfox98:
Change `mbedtls_sha256_update` and `mbedtls_sha256_finish` to `mbedtls_sha256_update_ret` and `mbedtls_sha256_finish_ret`, respectively. the non-`_ret` versions were depreciated in 2.7.0 according to the [documentation](https://mbed-ce.github.io/mbed-os/group__mbedtls__sha256__module.html).

I get a `error: void value not ignored as it ought to be` when trying to compile with the deprecated functions. Switching to the _ret versions fixed this for me.

### DIFF
--- a/port/freertos/golioth_sys_freertos.c
+++ b/port/freertos/golioth_sys_freertos.c
@@ -234,7 +234,7 @@ enum golioth_status golioth_sys_sha256_update(golioth_sys_sha256_t sha_ctx,
     }
 
     mbedtls_sha256_context *hash = sha_ctx;
-    int err = mbedtls_sha256_update(hash, input, len);
+    int err = mbedtls_sha256_update_ret(hash, input, len);
     if (err)
     {
         return GOLIOTH_ERR_FAIL;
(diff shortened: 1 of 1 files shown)


sam-golioth:
Hi @atfox98! Thanks for the contribution. Beginning with MbedTLS v3.0.0, the variants of those functions that return `void` were [removed](https://github.com/Mbed-TLS/mbedtls/pull/4334) and subsequently the variants that return an integer value were [renamed](https://github.com/Mbed-TLS/mbedtls/pull/4629) to remove the `_ret` suffix. We recommend updating MbedTLS in your system to the latest version.

Our first-party ports to ESP-IDF and Zephyr are both using MbedTLS v3.6, the latest version. In reviewing this change, we've determined that these specific hashing functions should be part of the platform port rather than making the assumption that all ports that use FreeRTOS are also using MbedTLS. That should make it easier for you to override the hashing functions in your custom port, although we still recommend updating to the latest version of MbedTLS to ensure you have the latest safety and security enhancements.

Because our supported ports are using the latest version of MbedTLS, we'll be closing this Pull Request, but please post on our [Forum](https://forum.golioth.io/) if you have any further questions or need assistance with anything!

PR #702, opened 2024-12-10
fw_update: Stability improvements to reporting state/reason

szczys:
resolves https://github.com/golioth/firmware-issue-tracker/issues/736

### DIFF
--- a/src/fw_update.c
+++ b/src/fw_update.c
@@ -356,6 +356,7 @@ static void fw_update_thread(void *arg)
 
         block_retries.idx = 0;
         block_retries.count = 0;
+        bool block_retries_reported = false;
 
         while (1)
         {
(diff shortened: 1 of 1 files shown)


src/fw_update.c
```diff
@@ -75,26 +78,60 @@ static enum golioth_status fw_write_block_cb(const struct golioth_ota_component
...
-                                                               const char *package,
-                                                               const char *current_version,
-                                                               const char *target_version,
-                                                               int32_t timeout_s)
+enum golioth_status golioth_fw_update_report_state_sync(struct fw_update_component_context *ctx,
+                                                        enum golioth_ota_state state,
+                                                        enum golioth_ota_reason reason,
+                                                        bool report_component_name,
+                                                        bool report_current_version,
+                                                        bool report_target_version)
```
mniestroj, line 86: How about replacing 3 `bool` arguments with single `int` argument that would allow flags to be ORed together with `REPORT_COMPONENT_NAME | REPORT_CURRENT_VERSION | REPORT_TARGET_VERSION`?

The problem with `bool` arguments is that when passing them to function, it is very hard to know what those are about. It requires to look at function prototype to see that each `true` or `false` does. On the other hand passing flags (macros) directly to function allows to see what exactly those parameters are about.

src/fw_update.c
```diff
@@ -75,26 +78,60 @@ static enum golioth_status fw_write_block_cb(const struct golioth_ota_component
...
-                                         reason,
-                                         package,
-                                         current_version,
-                                         target_version,
-                                         timeout_s);
+    enum golioth_status status;
+    uint32_t backoff_s = FW_REPORT_RETRIES_INITAL_DELAY_S;
+    int retries_remaining = FW_REPORT_MAX_RETRIES;
+
+retry_report:
```
mniestroj, line 102: Nit: it looks that replacing label+goto should be doable with `while (retries_remaining--)` logic. Behavior should be the same, but it will  probably be a bit easier to read due to widely used while/for loops compared to goto (with the main exception being error handling).

szczys: Thanks for noticing. This is much better as a while loop.

src/fw_update.c
```diff
@@ -182,16 +182,22 @@ static bool received_new_target_component(const struct golioth_ota_manifest *man
...
+         *  This can't be done directly because it would call a sync func from a callback
+         *  Consider adding a new reason code: GOLIOTH_OTA_REASON_COMPONENT_NOT_FOUND
+         */
+    }
+
     return false;
 }
 
-static void fw_report_and_observe(void)
+static void fw_register_observe(void)
```
sam-golioth, line 199: This function name feels a little awkward. Maybe `fw_register_manifest_observation()` or `fw_observe_manifest()`?

sam-golioth (approved):
Awesome! Love to see the \~synergies\~ between this and the work being done on OTA observability on the cloud.

PR #708, opened 2024-12-11
WiFi location service with sample

mniestroj:


include/golioth/location/wifi_zephyr.h
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <zephyr/net/wifi_mgmt.h>
+#include <golioth/location/wifi.h>
+
+static inline enum golioth_status golioth_location_wifi_append_zephyr(
```
hasheddan, line 12: I believe this would be the first platform-specific public header in the SDK. Given the relatively minimal amount of functionality it offers, do you think it is necessary to include in the SDK, rather than just showing how this can be done in samples?

include/golioth/location/wifi.h
```diff
@@ -0,0 +1,42 @@
...
+};
+
+void golioth_location_wifi_init(struct golioth_location_wifi_req *req);
+enum golioth_status golioth_location_wifi_append(struct golioth_location_wifi_req *req,
+                                                 const struct golioth_wifi_scan_result *entry);
+enum golioth_status golioth_location_wifi_finish(struct golioth_location_wifi_req *req);
+
+enum golioth_status golioth_location_wifi_get_sync(struct golioth_client *client,
+                                                   struct golioth_location_wifi_req *req,
+                                                   struct golioth_location_rsp *rsp);
```
hasheddan, line 42: For devices that support both Wi-Fi and cellular we will allow for both types of network data to be included in the same request. I could see value in having dedicated functions for each, but we could also have a single `golioth_location_get_sync` that accepted optional Wi-Fi and cellular requests. Doing so would make it harder enable / disable individually though. What do you think?

include/golioth/location/wifi_zephyr.h
mniestroj: Agree, that this is very minimal and can be used in samples directly.

include/golioth/location/wifi.h
mniestroj: Makes sense! I somehow missed the case with both technologies used at once. I'll rework current implementation to account for that.

src/coap_client.c
```diff
@@ -261,6 +261,34 @@ static enum golioth_status golioth_coap_client_set_internal(struct golioth_clien
     return GOLIOTH_OK;
 }
 
+enum golioth_status golioth_coap_client_post(struct golioth_client *client,
```
szczys, line 264: Do we have anywhere else that we use `_post` in the user API? I ask because this is a departure from the `_set` and `_set_block` API naming we currently have. I wonder if `_set_payload` or something along those lines would be a good fit?

src/coap_client.c
```diff
@@ -261,6 +261,34 @@ static enum golioth_status golioth_coap_client_set_internal(struct golioth_clien
...
+                                             golioth_post_cb_fn callback,
+                                             void *callback_arg,
+                                             bool is_synchronous,
+                                             int32_t timeout_s)
+{
+    struct golioth_coap_post_params params = {
+        .content_type = content_type,
+        .callback_post = callback,
+        .arg = callback_arg,
+        .callback_is_post = true,
```
szczys, line 279: Perhaps not relevant to this PR, but this `callback_is_post` member makes me think of the `is_observe` member we have in some other params structs. Perhaps these could be unified in a `type` member that all params structs have with an enum that specifies is_post, is_observe, etc.

include/golioth/location.h
```diff
@@ -0,0 +1,48 @@
...
+    /** Latitudal position in nanodegrees (0 to +-180E9) */
+    int64_t latitude;
+    /** Longitudal position in nanodegrees (0 to +-180E9) */
+    int64_t longitude;
+
+    /** Accuracy in meters */
+    int64_t accuracy;
+};
+
+#define GOLIOTH_LOCATION_FLAG_WIFI BIT(0)
```
szczys, line 25: is `BIT()` in the C standard library or is this Zephyr-specific?

src/coap_client.c
mniestroj: Yeah... I was thinking about that myself and considered `set_payload` / `set_with_response` / `set_with_rsp`, etc.

1. My first implementation was about extending `set` with payload (and payload_size) arguments. This seemed to be obvious choice for internals of our SDK and was the cleanest path forward, without introducing new API (like _post). This turned out to be very verbose for public API that SDK users consume. So... decided to leave those APIs **without** payload.
2. Next approach was to implement either `_set()` equivalent with `_set_with_rsp()` to describe that payload is provided in that case. That however was very loosely related to our existing `_set()` APIs, like `lightdb_*_set()`. Then I was thinking about whether `_set_*()` is a good choice for API that sends and receives something. More natural is the `_post()` nomenclature, which is very well known in case of HTTP, CoAP and in general in the world of RESTful APIs.
3. Decided to go with `_post()` API, since that maps very good to internal CoAP implementation without introducing new nomenclature. It does what POST method does technically... so it sends data and receives data. It is definitely more than **set** word describes.

So if we are about changing the (internal) APIs, my bet would be to use `_post()` in most places and just expose `_set()` for high-level APIs, like lightdb, where it makes sense. Even for **stream** service which just sends data now... I think it makes more sense to use `golioth_stream__push()` or `golioth_stream__post()` instead of `golioth_stream__set()`.

src/coap_client.c
mniestroj: Sure, I would 100% agree if we would have them in the same structure. Do you think it makes sense to introduce type now even if we have 2 types at most for each struct?

Even if we would switch to types, then I think that `observe` and `post` should be defined in separate enum, i.e. separate namespace. They are just different things, so I would prefer to not put them in single location to prevent confusion.

include/golioth/location.h
mniestroj: Good catch! Hmmm, that is Zephyr/Linux kernel specific... but hey... it was built successfully for all platforms I think. Anyway, we can switch to `(1 << 0)` just in case.

mniestroj: Well, it was not built, as there is only Zephyr example. So yeah, good point. Will stich to bit shift directly.

szczys (approved):

include/golioth/location.h
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (c) 2024 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
```
hasheddan, line 7: Let's go ahead and add our ifdef C++ macros here to match other include files 👍🏻

include/golioth/location.h
```diff
@@ -0,0 +1,48 @@
...
+    int flags;
+
+    struct golioth_location_rsp *rsp;
+#endif
+};
+
+void golioth_location_init(struct golioth_location_req *req);
+enum golioth_status golioth_location_finish(struct golioth_location_req *req);
+
+enum golioth_status golioth_location_get_sync(struct golioth_client *client,
```
hasheddan, line 46: Can we add doxygen comments on these?

src/coap_client.c
hasheddan: I also like using `_post` because it more closely matches our use of `_get` and `_delete` elsewhere. I don't love the idea of us now exposing a `set_cb` and a `post_cb` as part of the public API though. The `post_cb` looks very similar to the `get_cb` (identical?) and the `set_cb` is currently used for POST and DELETE operations. I wonder if moving to something more like "callback with response" and "callback without response" would make more sense?

examples/zephyr/location/wifi/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: Apache-2.0
```
hasheddan, line 1: It would be good to have a `README.md` for this example, especially given that this service is currently in private access. An explanation regarding the Wi-Fi playback functionality would likely also be useful 👍🏻

include/golioth/location.h
```diff
@@ -0,0 +1,48 @@
...
+#pragma once
+
+#include <golioth/client.h>
+#include <golioth/golioth_status.h>
+#include <stdint.h>
+#include <zcbor_encode.h>
+
+struct golioth_location_rsp
+{
+    /** Latitudal position in nanodegrees (0 to +-180E9) */
```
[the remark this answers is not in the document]
mniestroj: Hmm, does it mean that there is an error in Zephyr as well? https://github.com/zephyrproject-rtos/zephyr/blob/d789f339efb78a136a10c74efdeeb963e085e104/include/zephyr/data/navigation.h#L26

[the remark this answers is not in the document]
hasheddan: I believe so 😅

src/coap_client.c
sam-golioth: This is a good discussion! Here are my thoughts:

- I like the idea of using `_post()` for the **internal** API. The use of `set()` here always felt a little odd here, since we are doing POST operations and the other APIs also use the CoAP REST operation names (GET, OBSERVE, DELETE, etc.). We can keep `set()` for the higher level APIs that don't return a payload like LightDB State and Stream. This change would be out of scope for this PR.
- For this PR, I think adding the new API works fine.
- I agree with Dan that we have a lot of overlap in the callbacks. I'm working on a [change](https://github.com/golioth/golioth-firmware-sdk/compare/main...proxy_callbacks_thru_upper_client) to proxy the callbacks through the upper half client as part of my hackday / holiday hours project. For that change, I'm using one catch-all callback between the lower and upper-half clients. We could do something similar for user callbacks (i.e. use one callback and set the payload to NULL if it is irrelevant for the operation). Or we could rename the callbacks as Dan suggested. Another idea is proxying all callbacks through the higher level services and then using service-specific callbacks. I think that will be easier for users, but it does add a lot of indirection to the callbacks (lower client => upper client => service => user code).

sam-golioth (commented):
Great work @mniestroj! Can we add a twister test for the new sample? I'd like us to get into the habit of adding tests alongside new code rather than adding them retroactively.

include/golioth/location.h
```diff
@@ -0,0 +1,48 @@
...
+    int64_t latitude;
+    /** Longitudal position in nanodegrees (0 to +-180E9) */
```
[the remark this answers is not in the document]
mniestroj: Fixed

include/golioth/location.h
mniestroj: Addressed!

include/golioth/location/wifi.h
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) 2024 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
```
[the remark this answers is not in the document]
mniestroj: Doxygen comments were added

include/golioth/location.h
mniestroj: Doxygen comments were added

examples/zephyr/location/wifi/CMakeLists.txt
mniestroj: Readme file was added.

src/Kconfig
```diff
@@ -78,6 +78,25 @@ config GOLIOTH_LIGHTDB_STATE_FLOAT_HELPERS
 
 endif # GOLIOTH_LIGHTDB_STATE
 
+config GOLIOTH_LOCATION
+    bool "Golioth Location service"
+
+if GOLIOTH_LOCATION
+
+config GOLIOTH_LOCATION_REQUEST_BUFFER_SIZE
```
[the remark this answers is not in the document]
mniestroj: Addressed!

include/golioth/location.h
```diff
@@ -0,0 +1,87 @@
...
+ */
+struct golioth_location_req
+{
+#ifdef CONFIG_GOLIOTH_LOCATION
+    uint8_t buf[CONFIG_GOLIOTH_LOCATION_REQUEST_BUFFER_SIZE];
+    zcbor_state_t zse[2 /* backup */ + 1 /* WiFi only */];
+    enum golioth_status status;
+    int flags;
+
+    struct golioth_location_rsp *rsp;
```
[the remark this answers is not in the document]
mniestroj: Great feedback, agree with all above!

examples/zephyr/location/wifi/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
...
```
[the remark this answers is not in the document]
mniestroj: I was convinced that I have updated this commit message before... probably dropped the change after rebase and parallel work on cellular location. Anyway, thanks for spotting it!

include/golioth/location.h
```diff
@@ -0,0 +1,87 @@
...
+
+#define GOLIOTH_LOCATION_FLAG_WIFI (1 << 0)
+
+/**
+ * @brief Wi-Fi location request, used with golioth_location_init(), golioth_location_finish() and
+ *        golioth_location_get_sync()
+ */
+struct golioth_location_req
+{
+#ifdef CONFIG_GOLIOTH_LOCATION
```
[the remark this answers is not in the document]
mniestroj: I was **probably** `CONFIG_GOLIOTH_LOCATION_REQUEST_SIZE` as an error... but not sure now. I think I saw this pattern somewhere, but maybe it was just old SDK that I was grepping instead. Now dropped it and no issues in CI I think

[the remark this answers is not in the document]
sam-golioth: Maybe there was error with Linux or Modus before these definitions were added to `config.h`.

src/location.c
```diff
@@ -0,0 +1,139 @@
...
+LOG_TAG_DEFINE(location);
+
+struct golioth_location_get_sync_data
+{
+    enum golioth_status status;
+    const struct golioth_location_req *req;
+    struct golioth_location_rsp *rsp;
+};
+
+static enum golioth_status location_decode(const struct golioth_location_req *req,
```
sam-golioth, line 24: `req` is unused in this function.

src/location.c
```diff
@@ -0,0 +1,139 @@
...
+#include "coap_client.h"
+
+#if defined(CONFIG_GOLIOTH_LOCATION)
+
+LOG_TAG_DEFINE(location);
+
+struct golioth_location_get_sync_data
+{
+    enum golioth_status status;
+    const struct golioth_location_req *req;
```
sam-golioth, line 20: Do we need `req` in this struct?

sam-golioth (commented):
Changes look good @mniestroj! Left a couple comments about using the location request struct while processing the response. Still missing a twister test for the sample as well.

src/location.c
mniestroj: Removed it

src/location.c
mniestroj: Good catch, removed it as it was not needed anywhere really.

mniestroj:
> Still missing a twister test for the sample as well.

Working on that :) Which platforms do we want to support? I'm targeting `native_sim` for now, due to possiiblity to test with playback data. For real WiFi capable platforms we don't really have a reliable way of testing with accesspoint.

sam-golioth:
> > Still missing a twister test for the sample as well.
> 
> Working on that :) Which platforms do we want to support? I'm targeting `native_sim` for now, due to possiiblity to test with playback data. For real WiFi capable platforms we don't really have a reliable way of testing with accesspoint.

I'm good with just native_sim to start 👍

mniestroj:
Twister sample test implemented in #725 (it requires dev environment for now due to Location backend service).

sam-golioth (approved):
Awesome work @mniestroj!

PR #711, opened 2024-12-18
fw_update: add package name Kconfig symbol

szczys:
Add Kconfig symbol: CONFIG_GOLIOTH_FW_UPDATE_PACKAGE_NAME

Previously the package name used by the fw_update module was hardcoded in the header file. Make this value configurable so it may be used programmatically at build time.


src/Kconfig
```diff
@@ -43,6 +43,13 @@ config GOLIOTH_FW_UPDATE
 
 if GOLIOTH_FW_UPDATE
 
+config GOLIOTH_FW_UPDATE_PACKAGE_NAME
+    string "Package name of the main firmware component"
+    default "main"
+    help
+      The Package Name used by the Golioth Cloud to indicate the binary to use for device firmware
+      update. This name must be no more than GOLIOTH_OTA_MAX_PACKAGE_NAME_LEN characters in length.
```
sam-golioth, line 51: Let's add a static assert to enforce this

szczys: Great suggestion, thanks!

image

mniestroj (approved):

PR #712, opened 2024-12-20
workflows: add Clean CI Project and run once per day

szczys:
Add a workflow that finds and remove Devices, Certificates, Cohorts, Release, and Tags that are more that 1 day old. This workflow runs once per day.

## Sample output:

This test was run with `--days-old 0.001` (about 90 seconds) but the default in this commit is 1 day.

image

Resolves: https://github.com/golioth/firmware-issue-tracker/issues/740

### DIFF
--- a/scripts/ci/clean_ci_project
+++ b/scripts/ci/clean_ci_project
@@ -1,0 +1,202 @@
+#!/usr/bin/env python3
+
+__author__ = "Golioth, Inc."
+__copyright__ = "Copyright (c) 2024 Golioth, Inc."
+__license__ = "Apache-2.0"
+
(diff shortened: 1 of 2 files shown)


scripts/ci/clean_ci_project
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python3
+
+__author__ = "Golioth, Inc."
+__copyright__ = "Copyright (c) 2024 Golioth, Inc."
```
sam-golioth, line 4: Better merge it quick!

sam-golioth (approved):

PR #713, opened 2024-12-20
Run regular integration tests

szczys:
- Adapt Allure Report workflow to accept a custom string used to group reports
- Add nsim/linux integration tests and nsim zephyr sample tests every 4 hours and output their Allure Report to 'recurring' so the history is separate from main and branch tests.

Output of these test is available here: https://golioth.github.io/allure-reports/recurring

Depends on #714 
Resolves https://github.com/golioth/firmware-issue-tracker/issues/754

### DIFF
--- a/.github/workflows/ci-clean-project.yml
+++ b/.github/workflows/ci-clean-project.yml
@@ -1,4 +1,4 @@
-name: Clean CI Project
+name: "CI: Clean Project"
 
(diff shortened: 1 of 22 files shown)


hasheddan (approved):
Nice work @szczys!

sam-golioth (approved):

PR #716, opened 2024-12-31
workflows: standardize titles and filenames

szczys:
Rename and retitle all tests for better organization. This delivers a more uniform list on the GitHub Actions page, and groups workflow files by functionality.

- Use categories as prefix: ci, doc, hil, report, test, trigger
- Normalize on '-' instead of '_' in filenames
- Replace 'test' in filenames with 'integration' to differentiate from 'sample'

### DIFF
--- a/.github/workflows/ci-clean-project.yml
+++ b/.github/workflows/ci-clean-project.yml
@@ -1,4 +1,4 @@
-name: Clean CI Project
+name: "CI: Clean Project"
 
 on:
   schedule:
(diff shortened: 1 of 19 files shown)


mniestroj (approved):

sam-golioth (approved):
Awesome, thanks for taking the initiative here @szczys! When this merges, we're probably going to need to update the required checks in the branch protection settings to match the new names.

PR #723, opened 2025-01-07
workflow: use specific version of allure-report-action

szczys:
Specify most recent minor version of action to ensure the fixes for links back to the workflow associated with a report is available.

## Before

2025-01-07-105013_3840x1080_scrot

## After

2025-01-07-105045_3840x1080_scrot


sam-golioth (approved):
Thanks for the before/after! Made reviewing much easier 👍

PR #724, opened 2025-01-08
Update ESP-IDF to v5.4

szczys:
SDK Changes Due to this Update:
- Guard LOG_COLOR* defines with #ifndef
    - Espressif changed the defines so they no longer exactly match those in the Golioth SDK. They also turned off log    color by default, citing that `idf.py monitor` now handles colorization.
- Guard IN6_IS_ADDR_V4MAPPED(a) with #ifndef
    - The macro was compared between the header in the Golioth Firmware SDK and the one in the esp-lwip repo and found to be functionally equivalent.

Resolves https://github.com/golioth/firmware-issue-tracker/issues/734

### DIFF
--- a/port/esp_idf/libcoap/include/coap_config_posix.h
+++ b/port/esp_idf/libcoap/include/coap_config_posix.h
@@ -40,9 +40,12 @@ struct in6_pktinfo {
     struct in6_addr ipi6_addr; /* src/dst IPv6 address */
     unsigned int ipi6_ifindex; /* send/recv interface index */
 };
+
+#ifndef IN6_IS_ADDR_V4MAPPED
 #define IN6_IS_ADDR_V4MAPPED(a) \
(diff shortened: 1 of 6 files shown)


port/esp_idf/libcoap/include/coap_config_posix.h
```diff
@@ -40,9 +40,12 @@ struct in6_pktinfo {
     struct in6_addr ipi6_addr; /* src/dst IPv6 address */
     unsigned int ipi6_ifindex; /* send/recv interface index */
 };
+
+#ifndef IN6_IS_ADDR_V4MAPPED
```
sam-golioth, line 44: I appreciate the explanation in the commit message! I was going to ask why we didn't just remove it. But it looks like you the last bit of the message is missing.

sam-golioth (approved):
Thanks @szczys!

PR #725, opened 2025-01-10
Location WiFi - twister+pytest test for native_sim and playback data

mniestroj:


examples/zephyr/location/wifi/pytest/test_sample.py
```diff
@@ -0,0 +1,39 @@
+from collections import namedtuple
```
sam-golioth, line 1: Let's add a copyright header

sam-golioth (commented):
Nice work @mniestroj! The failures in prod will be annoying but I think we should still merge this in.

examples/zephyr/location/wifi/pytest/test_sample.py
mniestroj: Added in both `*.py` files.

sam-golioth (approved):
Oops! I thought I hit approve last week.

szczys (approved):

sam-golioth:
> The major benefit of having tests merged and running in the CI is to get feedback whenever something breaks... this won't be the case if we merge this one.

That's a great point. I'm good with waiting for the backend support 👍

mniestroj:
Closing in favor of #730, which has combined cellular and wifi tests.

PR #730, opened 2025-01-16
location: twister+pytest test for native_sim and playback data

mniestroj:
There are 3 runtime tests:
 * cellular + wifi
 * cellular only
 * wifi

all executed using `native_sim` platform and playback data.

Supersedes: #725

### DIFF
--- a/examples/zephyr/location/pytest/test_sample.py
+++ b/examples/zephyr/location/pytest/test_sample.py
@@ -1,0 +1,43 @@
+# Copyright (c) 2025 Golioth, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from collections import namedtuple
+import re
+
+from twister_harness.device.device_adapter import DeviceAdapter
+from twister_harness.helpers.shell import Shell
+
+import pytest
+
+
+pytestmark = pytest.mark.anyio
+
+
+async def test_location(shell: Shell, dut: DeviceAdapter, device):
+    # Set Golioth credential
+
+    golioth_cred = (await device.credentials.list())[0]
+    shell.exec_command(f"settings set golioth/psk-id {golioth_cred.identity}")
+    shell.exec_command(f"settings set golioth/psk {golioth_cred.key}")
+
+    # Wait for Golioth connection
+
+    dut.readlines_until(regex=".*Golioth CoAP client connected", timeout=90.0)
+
+    # Verify position
+
+    pattern = re.compile(r".* (?P<lon>\d+\.\d+) (?P<lat>\d+\.\d+) \((?P<acc>\d+)\)")
+    Position = namedtuple('Position', ('lon', 'lat', 'acc'))
+
+    positions = []
+    for _ in range(3):
+        lines = dut.readlines_until(regex=pattern, timeout=30.0)
+        m = pattern.search(lines[-1])
+        pos = Position(*[float(m[p]) for p in ['lon', 'lat', 'acc']])
+        positions.append(pos)
+
+    for pos in positions:
+        assert pos.lon == pytest.approx(50.663974800, 1e-3)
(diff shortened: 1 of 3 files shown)


examples/zephyr/location/sample.yaml
```diff
@@ -2,29 +2,47 @@ sample:
...
       - native_sim/native/64
-  sample.golioth.location.cellular_only:
+  sample.golioth.location.playback.cellular:
+    harness: pytest
+    extra_args: EXTRA_CONF_FILE="../common/runtime_settings.conf"
     extra_configs:
-      - CONFIG_GOLIOTH_WIFI_PLAYBACK=n
+      - CONFIG_GOLIOTH_SAMPLE_TWISTER_TEST=y
+      - arch:posix:CONFIG_NATIVE_UART_0_ON_STDINOUT=y
+      - CONFIG_GOLIOTH_CELLULAR_PLAYBACK=n
```
sam-golioth, line 45:
```suggestion
      - CONFIG_GOLIOTH_WIFI_PLAYBACK=n
```

mniestroj: Good catch. Updated and retested locally with dev:
```
INFO    - 1/4 native_sim/native/64      sample.golioth.location                            NOT RUN (build)
INFO    -                                    sample.golioth.location                                                     NOT RUN      Test was built only
INFO    - 2/4 native_sim/native/64      sample.golioth.location.playback.wifi              PASSED (native 12.779s)
INFO    -                                    sample.golioth.location.playback.wifi.test_location                         PASSED      
INFO    - 3/4 native_sim/native/64      sample.golioth.location.playback.cellular          PASSED (native 27.618s)
INFO    -                                    sample.golioth.location.playback.cellular.test_location                     PASSED      
INFO    - 4/4 native_sim/native/64      sample.golioth.location.playback                   PASSED (native 37.775s)
INFO    -                                    sample.golioth.location.playback.test_location                              PASSED      

INFO    - 4 test scenarios (4 configurations) selected, 0 configurations filtered (0 by static filter, 0 at runtime).
```

sam-golioth (approved):

examples/zephyr/location/pytest/test_sample.py
```diff
@@ -0,0 +1,43 @@
...
+    for _ in range(3):
+        lines = dut.readlines_until(regex=pattern, timeout=30.0)
+        m = pattern.search(lines[-1])
+        pos = Position(*[float(m[p]) for p in ['lon', 'lat', 'acc']])
+        positions.append(pos)
+
+    for pos in positions:
+        assert pos.lon == pytest.approx(50.663974800, 1e-3)
+        assert pos.lat == pytest.approx(17.942322850, 1e-3)
+        assert pos.acc < 1000
```
hasheddan, line 43: After a successful check of the three resolved positions will this terminate the program early? If we run this on every PR I'd prefer to avoid making a large number of location requests each time 👍🏻

mniestroj: Yes, `native_sim` process gets killed as soon as the test finished. This means right after checking those 3 positions.

hasheddan (approved):

examples/zephyr/location/pytest/test_sample.py
hasheddan, line 43: Per offline conversation, it likely makes sense to remove this check, replace it with an existence check (e.g. `acc != 0`), or increase the HPE tolerance. While accuracy is important, testing the accuracy value of the location service is not the intended purpose of this test.

mniestroj: Updated to 2000 for now, just to make sure we have a reasonable value returned and properly parsed on the firmware side. We can look into removing this check or adjusting in the future at any point.

PR #743, opened 2025-02-03
fw_update: check for existing candidate image before attempting download

hasheddan:
Refactors `fw_update` logic to check whether the existing image in the candidate slot matches the desired update image in a received manifest. If so, we attempt to boot into the image without downloading it again.

Implementing involved some cleanup of the interface to `fw_update` platform ports. Namely, `fw_update_post_download` now returns a status, and `fw_update_validate` receives an image hash and size. The former is to allow for ports to return an error if they are unable to "finalize" a download. This is particularly relevant for `esp_idf`, as completing the download also implicitly performs image validation. The latter change is to allow for validation to be performed on a candidate image outside the context of the download of that image, which may have taken place prior to the most recent reboot.

Because ports can perform custom validation in `fw_update_post_download` _and_ we implement hash validation in `fw_update` for downloaded images, the `fw_update_validate` implementation for a port is no longer invoked after download. Instead, it is now only invoked on receiving a new manifest to check whether the desired update image is already present. The primary reason for not also invoking `fw_update_validate` post download is that it is not implemented for all ports at this time, though it is also somewhat redundant with our existing hash check. There are two options that would allow for calling `fw_update_validate` post download _and_ prior to download:

1. Provide implementations for all ports.
2. Allow for differentiating between "not supported" and "validation failed".

(1) feels like the right path long term, and given that we do not lose any functionality today by removing the post download call (i.e. ports previously just had no-op implementations), moving forward with the `fw_update_post_download` + `fw_update`'s hash check seems like a reasonable course of action.

---

Work items that should follow this implementation:
- We should update our OTA HIL testing to include downloading an image that will never connect to Golioth, then ensure that when we rollback and see the manifest again that we attempt to boot into it without download. The current tests effectively just validate that `fw_update_validate` always fails.
- This implementation avoids downloading the same image over and over again, but we should also implement some sort of upgrade backoff to avoid booting into the new image over and over again. Continuously doing so results in flash wear from swapping the candidate image into the primary slot on upgrade and revert.


port/zephyr/golioth_fw_zephyr.c
```diff
@@ -221,25 +242,90 @@ enum golioth_status fw_update_handle_block(const uint8_t *block,
...
     }
+
+    return GOLIOTH_OK;
 }
 
-enum golioth_status fw_update_validate(void)
+/*
+ * @note This is similar to Zephyr's flash_img_check() but uses the Golioth
+ * port's sha256 API.
+ */
```
hasheddan, line 267: In addition to being consistent with our sha256 implementation, this is also necessary because building with `flash_img_check` enabled with mbedtls is currently broken on nordic platforms due to https://github.com/zephyrproject-rtos/zephyr/pull/85238

port/linux/fw_update_linux.c
```diff
@@ -85,9 +85,10 @@ enum golioth_status fw_update_post_download(void)
     return GOLIOTH_OK;
 }
 
-enum golioth_status fw_update_validate(void)
+enum golioth_status fw_update_validate(const uint8_t *hash, const size_t size)
 {
-    return GOLIOTH_OK;
+    // TODO(hasheddan): support validating candidate firmware.
+    return GOLIOTH_ERR_FAIL;
```
szczys, line 91:
``` suggestion
    return GOLIOTH_ERR_NOT_IMPLEMENTED;
```

port/modus_toolbox/fw_update_mcuboot.c
```diff
@@ -116,10 +116,10 @@ enum golioth_status fw_update_post_download(void)
     return GOLIOTH_OK;
 }
 
-enum golioth_status fw_update_validate(void)
+enum golioth_status fw_update_validate(const uint8_t *hash, const size_t size)
 {
-    // Nothing to do
-    return GOLIOTH_OK;
+    // TODO(hasheddan): support validating candidate firmware.
+    return GOLIOTH_ERR_FAIL;
```
szczys, line 122:
``` suggestion
    return GOLIOTH_ERR_NOT_IMPLEMENTED;
```

port/zephyr/golioth_fw_zephyr.c
```diff
@@ -240,9 +261,71 @@ enum golioth_status fw_update_post_download(void)
...
+    err = flash_area_open(UPLOAD_FLASH_AREA_ID,
+                          (const struct flash_area **) &(_flash_img_context.flash_area));
+    if (err)
+    {
+        return GOLIOTH_ERR_IO;
+    };
+
+    golioth_sys_sha256_t hash_ctx = golioth_sys_sha256_create();
+
+    buf_size = sizeof(_flash_img_context.buf);
```
szczys, line 285: Note for other reviewers: This buf size defaults to `CONFIG_IMG_BLOCK_BUF_SIZE` which defaults to 512.

port/esp_idf/fw_update_esp_idf.c
```diff
@@ -104,8 +105,28 @@ enum golioth_status fw_update_post_download(void)
...
+    };
+    if (esp_image_verify(ESP_IMAGE_VERIFY, &part_pos, &data) != ESP_OK)
+    {
+        return GOLIOTH_ERR_FAIL;
+    }
+    if (data.image_len != size)
+    {
+        return GOLIOTH_ERR_FAIL;
+    }
+    if (memcmp(hash, data.image_digest, sizeof(data.image_digest)) != 0)
```
szczys, line 125: Do we need to verify the length of `hash` is the same as the length of `data.image_digest`?

szczys (commented):
Great work @hasheddan!

port/esp_idf/fw_update_esp_idf.c
hasheddan: We don't know the length of `hash` here -- it is somewhat implicit in the port API contract as implemented, so it may be worth making it more explicit that a 32 byte sha256 hash is being passed (at the very least in the documentation on `fw_update_validate`). In practice, because `data.image_digest` is [hard-coded to 32 bytes](https://github.com/espressif/esp-idf/blob/c5865270b50529cd32353f588d8a917d89f3dba4/components/bootloader_support/include/esp_image_format.h#L35) and we always pass a `GOLIOTH_OTA_COMPONENT_BIN_HASH_LEN` (32) then they should always be same length.

I am still in process of testing the esp-idf `fw_update_validate` implementation though, so there may need to be some further changes made here.

szczys: Ah, thanks for the background.

>In practice, because data.image_digest is hard-coded to 32 bytes and we always pass a GOLIOTH_OTA_COMPONENT_BIN_HASH_LEN (32) then they should always be same length.

Would it make sense to change the function as below?

``` c
enum golioth_status fw_update_validate(const uint8_t hash[GOLIOTH_OTA_COMPONENT_BIN_HASH_LEN],
                                       const size_t size)
```

hasheddan: I typically shy away from using arrays as function parameters in C because in practice they decay to a pointers. However, they can give the impression that, for example, `sizeof(hash)` in a function implementation would return the size of the array, but in reality it would return the size of the pointer. Some folks certainly argue the other side of this, so I'm open to discussion 🙂

szczys (approved):

sam-golioth:
> Because ports can perform custom validation in `fw_update_post_download` _and_ we implement hash validation in `fw_update` for downloaded images, the `fw_update_validate` implementation for a port is no longer invoked after download. Instead, it is now only invoked on receiving a new manifest to check whether the desired update image is already present. The primary reason for not also invoking `fw_update_validate` post download is that it is not implemented for all ports at this time, though it is also somewhat redundant with our existing hash check. There are two options that would allow for calling `fw_update_validate` post download _and_ prior to download:
> 
> 1. Provide implementations for all ports.
> 2. Allow for differentiating between "not supported" and "validation failed".
> 
> (1) feels like the right path long term, and given that we do not lose any functionality today by removing the post download call (i.e. ports previously just had no-op implementations), moving forward with the `fw_update_post_download` + `fw_update`'s hash check seems like a reasonable course of action.

I think removing `fw_update_validate()` from the post-download flow (along with the change to `fw_update_post_download()`) is actually the right long term move. The only thing we do between invoking `fw_update_post_download()` and `fw_update_validate()` is verify the SHA hash (as you point out), and that shouldn't affect the port's ability to validate the image. Having two callbacks there is actually somewhat confusing, and doesn't add anything. In the future, if we implement patching or compression, it could make sense to have hooks before and after the post-download processing steps, but we shouldn't be trying to pre-design that.

Given the above, it could make sense to rename `fw_update_validate()` to make its new purpose clearer.

include/golioth/fw_update.h
```diff
@@ -121,15 +121,17 @@ enum golioth_status fw_update_handle_block(const uint8_t *block,
...
-/// attempting to boot into it.
+/// Called by golioth_fw_update.c prior to downloading a new image
+/// to check whether it is already present in candidate slot.
+///
+/// @param hash Image SHA256 hash of size GOLIOTH_OTA_COMPONENT_BIN_HASH_LEN
+/// @param size Image size, in bytes
 ///
 /// @return GOLIOTH_OK - image validated
 /// @return Otherwise - error in validation, abort firmware update
-enum golioth_status fw_update_validate(void);
```
sam-golioth, line 132: This commit (and the next four) won't build without the following five commits. We should preserve bisectability by either squashing these commits or just updating the function signatures in this commit (without the functional changes).

src/fw_update.c
```diff
@@ -424,6 +453,18 @@ static void fw_update_thread(void *arg)
                 continue;
             }
 
+            if (fw_update_validate(_component_ctx.target_component.hash,
+                                   _component_ctx.target_component.size)
+                == GOLIOTH_OK)
```
sam-golioth, line 458: I'm sure this is what `clang-format` requires, but I'm not a fan haha. If you want to [disable formatting](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code) for this line and replace it with something cleaner, I wouldn't be opposed.

src/fw_update.c
```diff
@@ -604,14 +630,11 @@ static void fw_update_thread(void *arg)
         /* Download successful. Reset backoff */
         backoff_reset(&_component_ctx);
 
-        int countdown = 5;
-        while (countdown > 0)
+        if (fw_change_image_and_reboot() != GOLIOTH_OK)
```
sam-golioth, line 633: This reports the `DOWNLOADING` state to the cloud and changes the boot image, so we don't need to do that in lines 616 - 628 anymore.

sam-golioth (commented):
Great work @hasheddan! Overall looks good, left a few comments

src/fw_update.c
hasheddan: I believe we still need the `DOWNLOADED` report, but agree we can remove the `UPDATING` and boot image swap 👍🏻

src/fw_update.c
```diff
@@ -544,6 +539,14 @@ static void fw_update_thread(void *arg)
             continue;
         }
 
+        if (GOLIOTH_OK != fw_update_post_download())
+        {
+            GLTH_LOGE(TAG, "Failed to perform post download operations");
+            fw_download_failed(GOLIOTH_OTA_REASON_FIRMWARE_UPDATE_FAILED);
+            golioth_sys_sha256_destroy(download_ctx.sha);
```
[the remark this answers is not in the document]
hasheddan: ack! was avoiding making that change in this PR, but would love to go ahead and do so -- I think it probably makes the most sense after the `err` check block (i.e. don't need to finish if we are going to destroy), but happy to always do finish + destroy after `back_increment` if you find it cleaner 👍🏻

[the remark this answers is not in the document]
sam-golioth: After the `err` check works for me!

src/fw_update.c
sam-golioth: Yep! I mean to write `UPDATING` - agreed we need to keep `DOWNLOADED`.

src/fw_update.c
```diff
@@ -289,25 +284,20 @@ static void fw_observe_manifest(void)
 }
 
 static enum golioth_status fw_verify_component_hash(
-    struct download_progress_context *ctx,
+    const uint8_t calc_sha256[GOLIOTH_OTA_COMPONENT_BIN_HASH_LEN],
     const uint8_t server_sha256[GOLIOTH_OTA_COMPONENT_BIN_HASH_LEN])
```
hasheddan, line 288: As mentioned earlier in review, I don't really love using arrays in function signatures, but given that is the existing pattern here, I opted to remain consistent.

sam-golioth: It's gives a false sense of compiler security, but from a readability standpoint I think it's the clearest way in C to communicate to the user that their input must be this exact size, no more and no less.

sam-golioth (approved):
🎉

PR #747, opened 2025-02-13
Implement multi-block upload

szczys:
Multi-block upload implements a blockwise upload using repeated API calls:
1. Create a context
2. Call the multi-block set API once for each block in the upload
3. Destroy the context

### DIFF
--- a/src/coap_blockwise.c
+++ b/src/coap_blockwise.c
@@ -306,6 +306,87 @@ finish:
     return status;
 }
 
+// Create an async upload context
+struct blockwise_transfer *golioth_blockwise_upload_start(struct golioth_client *client,
+                                                          const char *path_prefix,
+                                                          const char *path,
+                                                          enum golioth_content_type content_type)
+{
+    if (NULL == client || NULL == path_prefix || NULL == path)
+    {
+        return NULL;
+    }
+
+    if (strlen(path) > CONFIG_GOLIOTH_COAP_MAX_PATH_LEN)
+    {
+        GLTH_LOGE(TAG, "Path too long: %zu > %zu", strlen(path), CONFIG_GOLIOTH_COAP_MAX_PATH_LEN);
+        return NULL;
+    }
+
+    struct blockwise_transfer *ctx =
+        blockwise_transfer_init(client, path_prefix, path, content_type);
+    if (NULL == ctx)
+    {
+        return NULL;
+    }
+
+    /* Allocate and store path; freed in golioth_blockwise_upload_finish() */
+    char *path_buff = golioth_sys_malloc(strlen(path) + 1);
+    if (NULL == path_buff)
+    {
+        goto finish_with_ctx;
+    }
+
+    strncpy(path_buff, path, strlen(path) + 1);
+    ctx->path = path_buff;
+
+    return ctx;
+
+finish_with_ctx:
+    blockwise_transfer_free(ctx);
+    return NULL;
+}
+
+// Destroy an async upload context
+void golioth_blockwise_upload_finish(struct blockwise_transfer *ctx)
+{
+    /* ctx->path was allocated in golioth_blockwise_upload_start() */
+    free((char *) ctx->path);
+    blockwise_transfer_free(ctx);
+}
+
+// Send a single block asynchronously
+enum golioth_status golioth_blockwise_upload_block(struct blockwise_transfer *ctx,
+                                                   uint32_t block_idx,
+                                                   const uint8_t *block_buffer,
+                                                   size_t block_len,
+                                                   bool is_last,
+                                                   golioth_set_block_cb_fn set_cb,
+                                                   void *callback_arg,
+                                                   bool is_synchronous,
+                                                   int32_t timeout_s)
+{
+    if (NULL == ctx || NULL == block_buffer)
+    {
+        return GOLIOTH_ERR_NULL;
+    }
+
+    return golioth_coap_client_set_block(
+        ctx->client,
+        ctx->token,
+        ctx->path_prefix,
+        ctx->path,
+        is_last,
+        ctx->content_type,
+        block_idx,
+        BLOCKSIZE_TO_SZX(CONFIG_GOLIOTH_BLOCKWISE_UPLOAD_MAX_BLOCK_SIZE),
+        block_buffer,
+        block_len,
+        set_cb,
+        callback_arg,
+        is_synchronous,
+        timeout_s);
+}
 
 /* Blockwise Downloads related functions */
 
--- a/src/coap_blockwise.c
+++ b/src/coap_blockwise.c
@@ -137,7 +137,7 @@ static void on_block_sent(struct golioth_client *client,
                           enum golioth_status status,
                           const struct golioth_coap_rsp_code *coap_rsp_code,
                           const char *path,
-                          size_t block_szx,
+                          size_t block_size,
                           void *arg)
 {
     assert(arg);
@@ -145,7 +145,7 @@ static void on_block_sent(struct golioth_client *client,
     ctx->status = status;
     ctx->coap_rsp_code.code_class = coap_rsp_code->code_class;
     ctx->coap_rsp_code.code_detail = coap_rsp_code->code_detail;
-    ctx->negotiated_blocksize_szx = block_szx;
+    ctx->negotiated_blocksize_szx = BLOCKSIZE_TO_SZX(block_size);
 
     golioth_sys_sem_give(ctx->sem);
 }
--- a/src/coap_blockwise.c
+++ b/src/coap_blockwise.c
@@ -22,24 +22,15 @@ _Static_assert(BLOCKSIZE_TO_SZX(CONFIG_GOLIOTH_BLOCKWISE_UPLOAD_MAX_BLOCK_SIZE) 
 
 struct blockwise_transfer
 {
-    bool is_last;
+    struct golioth_client *client;
     enum golioth_content_type content_type;
     golioth_sys_sem_t sem;
     uint8_t token[GOLIOTH_COAP_TOKEN_LEN];
-    uint32_t block_idx;
-    /* Negotiated block size */
-    size_t block_size;
     const char *path_prefix;
     const char *path;
-    uint8_t *block_buffer;
-    union
-    {
-        read_block_cb read_cb;
-        write_block_cb write_cb;
-    } callback;
+
     enum golioth_status status;
     struct golioth_coap_rsp_code coap_rsp_code;
-    void *callback_arg;
 };
 
 struct post_block_ctx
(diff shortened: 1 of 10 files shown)


src/coap_blockwise.c
```diff
@@ -283,6 +311,72 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
...
+    struct blockwise_transfer_multi *ctx =
+        golioth_sys_malloc(sizeof(struct blockwise_transfer_multi));
+    if (!ctx)
+    {
+        return NULL;
+    }
+
+    golioth_coap_next_token(ctx->token);
+    memset(ctx->path, 0, sizeof(ctx->path));
+    memcpy(ctx->path, path, strlen(path));
```
sam-golioth, line 336: We can just use a single `strncpy()` call instead of these two calls.

src/coap_blockwise.c
```diff
@@ -283,6 +311,72 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
     return status;
 }
 
+// Create an async upload context
+struct blockwise_transfer_multi *golioth_blockwise_multi_upload_ctx_create(
```
sam-golioth, line 315: Can we increase code re-use between these new functions and the existing single shot API? The single shot API should be able to use these functions for part of its implementation.

src/coap_blockwise.c
```diff
@@ -39,6 +42,15 @@ struct blockwise_transfer
     void *callback_arg;
 };
 
+struct blockwise_transfer_multi
```
sam-golioth, line 45: Let's increase code re-use by including this in the context for the single shot API instead of duplicating fields. All of these fields are relevant to a blockwise transfer, whether the API is single shot or multipart.

sam-golioth, line 45: Since this is the public facing struct, let's keep it simple and clear for users, and rename the context struct for the single shot API.
```suggestion
struct blockwise_transfer
```

sam-golioth (changes requested):

src/coap_blockwise.c
szczys: This is a good call-out, thanks!

I added a commit that breaks out almost all initialization into the existing `blockwise_upload_init()` and also added a `blockwise_upload_free()`. This way, all block upload functions can use these two functions. 

There were a couple of choices made as part of this:
- The existing sync function guarantees `path` is not freed before the the block download completes, but this new multipart approach can't make that guarantee. I added `char *stored_path` as a context member and store the path there when needed
- I have a bit of angst about `path_prefix`. This is only used internally to the SDK and we pass a `#define` constant. (But that's not enforced or documented.) As long as we continue this practice and don't pass a prefix from the user API we avoid use-after-free issues. I left this as is.
- The two upload approaches use different callback types. I removed callback (and callback_arg) from the init function. These should be set by the caller (if desired) after initialization.

src/coap_blockwise.h
```diff
@@ -26,6 +28,24 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
...
+// Note: callback_arg will be the same for all blocks
+struct blockwise_transfer *golioth_blockwise_upload_start(const char *path_prefix,
+                                                          const char *path,
+                                                          enum golioth_content_type content_type,
+                                                          golioth_set_cb_fn cb,
+                                                          void *callback_arg);
+
+void golioth_blockwise_upload_stop(struct blockwise_transfer *ctx);
+
+enum golioth_status golioth_blockwise_upload_block(struct golioth_client *client,
```
sam-golioth, line 42: `client` is the same for all blocks, so it will be clearer to have it in `_start()` and store it in the context.

src/coap_blockwise.h
```diff
@@ -26,6 +28,24 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
...
+                                                   struct blockwise_transfer *ctx,
+                                                   uint32_t block_idx,
+                                                   uint8_t *block_buffer,
```
sam-golioth, line 45:
```suggestion
                                                   const uint8_t *block_buffer,
```

src/coap_blockwise.h
```diff
@@ -26,6 +28,24 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
...
+                                                   size_t data_len,
```
sam-golioth, line 46: Length arguments should match the naming of the corresponding pointer argument, for clarity.
```suggestion
                                                   size_t block_len,
```

src/coap_blockwise.h
```diff
@@ -26,6 +28,24 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
                                            golioth_set_cb_fn callback,
                                            void *callback_arg);
 
+/* Blockwise Multi-Part Upload */
+
+// Note: callback_arg will be the same for all blocks
+struct blockwise_transfer *golioth_blockwise_upload_start(const char *path_prefix,
+                                                          const char *path,
+                                                          enum golioth_content_type content_type,
+                                                          golioth_set_cb_fn cb,
```
sam-golioth, line 37: The callback function should be per-block.

src/coap_blockwise.c
```diff
@@ -28,11 +31,13 @@ struct blockwise_transfer
     size_t block_size;
     const char *path_prefix;
     const char *path;
+    char *stored_path;
```
sam-golioth, line 34: Instead of having separate members and separate codepaths, we can just use `path`. To ensure that the path remains valid, a copy can be allocated in `golioth_blockwise_upload_start()` and freed in `golioth_blockwise_upload_stop()`.

src/coap_blockwise.c
```diff
@@ -28,11 +31,13 @@ struct blockwise_transfer
...
     uint8_t *block_buffer;
     union
     {
         read_block_cb read_cb;
         write_block_cb write_cb;
+        golioth_set_cb_fn set_cb;
```
sam-golioth, line 40: I wasn't clear in my previous comment. We should not be allocating this full struct for all transfers. We should define two structs. The first one (`struct blockwise_transfer`) contains the information necessary to perform a blockwise transfer using the multi-part API. The second struct contains the information needed to execute the single-shot API. Because the second struct is a superset of the first struct, it should contain a `struct blockwise_transfer` as one of its members, rather than duplicating the fields.

src/coap_blockwise.c
```diff
@@ -59,25 +64,84 @@ struct get_block_ctx
...
+                                                 bool store_path)
 {
+    if ((store_path) && (strlen(path) > CONFIG_GOLIOTH_COAP_MAX_PATH_LEN))
+    {
+        GLTH_LOGE(TAG, "Path too long: %zu > %zu", strlen(path), CONFIG_GOLIOTH_COAP_MAX_PATH_LEN);
+        return NULL;
+    }
+
+    struct blockwise_transfer *ctx = golioth_sys_malloc(sizeof(struct blockwise_transfer));
+    if (!ctx)
```
sam-golioth, line 93: I haven't been strict about this in the past, but we should not use `!` for `NULL` comparisons. CERT recommends that the standalone `if (!x)` notation is [never used](https://wiki.sei.cmu.edu/confluence/display/c/EXP20-C.+Perform+explicit+tests+to+determine+success%2C+true+and+false%2C+and+equality). The explicit comparison is easier to read, and thus easier to maintain and less prone to bugs.
```suggestion
    if (NULL == ctx)
```

src/coap_blockwise.c
```diff
@@ -98,6 +162,22 @@ static void on_block_sent(struct golioth_client *client,
...
+                                size_t block_szx,
+                                void *arg)
+{
+    struct blockwise_transfer *ctx = (struct blockwise_transfer *) arg;
+
+    if (ctx->callback.set_cb)
+    {
+        ctx->callback.set_cb(client, status, coap_rsp_code, path, ctx->callback_arg);
+    }
+}
```
sam-golioth, line 179: Why proxy the callback through here? It appears we're only dropping the `block_szx` argument, but the user will need to know the negotiated blocksize if the server requests a different one. Though it could make sense to proxy through here in order to convert SZX to bytes.

src/coap_blockwise.c
```diff
@@ -59,25 +64,84 @@ struct get_block_ctx
...
-    ctx->callback_arg = callback_arg;
-    ctx->callback.read_cb = cb;
+    ctx->callback_arg = NULL;
+    ctx->callback.read_cb = NULL; /* sets all union members to NULL */
     golioth_coap_next_token(ctx->token);
+
+    return ctx;
+
+free_and_finish:
+    blockwise_upload_free(ctx);
```
sam-golioth, line 141: To avoid errors with partially initialized structures, it's best practice for the initializing function to clear its own memory on error. The free function then always operates on a fully allocated and initialized structure.

For example, if this initialization function fails to allocate the block buffer, then it will call `blockwise_upload_free()` which will attempt to destroy a semaphore set to NULL. But because the underlying semaphore implementation is eventfd, the result of that operation will be closing the file with descriptor `0`, which will likely succeed but destabilize the system.

Instead of calling `blockwise_upload_free()` here, we should use the chained goto cleanup method as before.

src/coap_blockwise.h
```diff
@@ -26,6 +28,24 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
...
+                                                          void *callback_arg);
+
+void golioth_blockwise_upload_stop(struct blockwise_transfer *ctx);
```
sam-golioth, line 40: After thinking about this some more, I think `finish` is clearer than `stop` here. `stop` indicates that an ongoing operation without a defined endpoint is being stopped, or an operation is being aborted before its completion. `finish` more accurately reflects that we have reached the conclusion of the operation.

sam-golioth (changes requested):

src/coap_blockwise.c
szczys: Ah, sorry for the confusion.

I've split several members out of `blockwise_transfer` and into the contexts used by sync get/post. I think it might be possible to combine get/post structs and use a unified init/free for those but that should be future work.

src/coap_blockwise.c
```diff
@@ -270,54 +291,113 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
...
+        ctx->path,
+        is_last,
+        ctx->content_type,
+        block_idx,
+        BLOCKSIZE_TO_SZX(CONFIG_GOLIOTH_BLOCKWISE_UPLOAD_MAX_BLOCK_SIZE),
+        block_buffer,
+        block_len,
+        set_cb,
+        callback_arg,
+        false,
```
sam-golioth, line 385: If we add `is_synchronous` and `timeout` arguments, then we can support synchronous operations here as well. The start and finish functions would be the same, so a pretty minimal change to gain that functionality.

sam-golioth (approved):
Nice work @szczys! This is much cleaner. Let's rebase to pick up the `trio-websockets` fix so we can the new tests pass 👍

PR #753, opened 2025-03-11
Bump NCS

mniestroj:


sam-golioth (approved):

sam-golioth:
@mniestroj we also need to update the revision in the Verified Devices table in `README.md`

PR #759, opened 2025-03-14
support PSA Crypto API by selecting proper Kconfig options

mniestroj:
This has surfaced when working with `native_sim` and Bluetooth. Bluetooth host pulls PSA crypto API as dependency. mbedTLS Kconfig options for PSA crypto API are different, so we were not selecting enough algorithms in order to connect to cloud. This PR adds those options, so that applications with `CONFIG_MBEDTLS_PSA_CRYPTO_C=y` will be covered when adding Golioth SDK as dependency.

DNM before bump to Zephyr 4.1.0, since it was not tested before. Bluetooth in 4.0.0 does not requre PSA crypto (it uses tinycrypt instead), but in 4.1.0 it does.


sam-golioth (approved):
Thanks @mniestroj! I especially appreciate the initiative to include tests here. The PSA changes look good, but I want to hold merging until we resolve the test failures on the RAK5010 related to reformatting the LFS partition.

PR #764, opened 2025-04-16
zephyr: drop QEMU as CVB

sam-golioth:


mniestroj:
There are `qemu_x86.conf` overlay files in `boards/` directories of each sample. Let's drop them.

PR #773, opened 2025-04-22
examples: zephyr: cert_provisioning: override LFS partition before boot

mniestroj:
Make sure to reformat LittleFS partition by overriding first 2 (erase)
blocks of the partition.


sam-golioth (commented):
Changes look good, but we need to figure out the RAK5010 failures

sam-golioth (approved):

PR #783, opened 2025-05-06
ci: remove 'twister-out' before extracting new artifacts

mniestroj:
In some cases there are still old artifacts on the filesystem on
self-hosted runners. They are no harm for testing, but they do occupy space
and they are archived as artifacts after HIL testing finishes.

Make sure that 'twister-out' is removed before extracting new artifacts.
This reduces confusion when inspecting failed jobs.

### DIFF
--- a/.ci-west-ncs.yml
+++ b/.ci-west-ncs.yml
@@ -14,7 +14,6 @@ manifest:
         - nrfxlib
         - oberon-psa-crypto
         - segger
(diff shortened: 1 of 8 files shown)


sam-golioth (approved):
Thanks @mniestroj! I've definitely been tripped up by this after extracting test artifacts and finding more runs/boards than expected.

PR #784, opened 2025-05-06
Gateway

sam-golioth:
Adds the Gateway service (uplink only for now) by applying the patches currently in the gateway repo.

### DIFF
--- a/include/golioth/gateway.h
+++ b/include/golioth/gateway.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2025 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#pragma once
+
+#include <golioth/golioth_status.h>
+#include <golioth/client.h>
+#include <stdlib.h>
+
+/// Start a gateway uplink. The pointer returned from this function can be used with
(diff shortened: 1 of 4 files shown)


include/golioth/gateway.h
```diff
@@ -0,0 +1,32 @@
+/*
```
hasheddan, line 1: Can we add doxygen comments and `extern C` ifdefs on this header?

src/gateway.c
```diff
@@ -0,0 +1,82 @@
+/*
```
hasheddan, line 1: Do we want to add a `CONFIG_GOLIOTH_GATEWAY` kconfig?

src/gateway.c
```diff
@@ -0,0 +1,41 @@
...
+ * Copyright (c) 2024 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <golioth/gateway.h>
+#include "coap_blockwise.h"
+
+#if defined(CONFIG_GOLIOTH_GATEWAY)
```
trond-snekvik, line 10: I lack context here, really, but how come we're doing this throughout instead of zephyr's cmake utils?

```cmake
zephyr_library_sources_ifdef(CONFIG_GOLIOTH_GATEWAY, src/gateway.c)
```

trond-snekvik (approved):

mniestroj (approved):

PR #788, opened 2025-05-15
ci: bump golioth-hil-base and switch to 'uv pip'

mniestroj:
This image has 'uv' installed, so it should be faster to install
dependencies with 'uv pip'.


sam-golioth (commented):
What about `hil-sample-zephyr.yml`?

mniestroj:
> What about `hil-sample-zephyr.yml`?

Can do it separately (next PR) or within this PR once https://github.com/golioth/fw-oci-images/pull/54 gets merged. In any case, this change is coming as well.

PR #796, opened 2025-06-04
Add blockwise get for OTA manifest

szczys:
This PR begins by moving the existing `golioth_get_block_cb_fn` typedef to the `coap_client.c` as this is the only place it is used. I do not consider this a breaking change as users would not have been able to do anything with this callback (unless they're calling private functions).

The `golioth_get_block_cb_fn` is then replaced and `golioth_end_block_cb_fn` is added. These match the format and param names of the private callbacks in `coap_blockwise.c`, so those callbacks are updated to use the public version.

The Zephyr port didn't support content type for blockwise get so that is implemented in this PR.

Finally, a very simple function for blockwise download of manifest files is added.


sam-golioth (approved):
Thanks @szczys! Just two small change requests.

PR #802, opened 2025-06-30
ci: bump Docker images to latest

mniestroj:
Latest Docker images come with Python 3.13 instead of Python 3.11. There is
also venv already setup, so that installed packages do not conflict with
system packages.

### DIFF
--- a/.github/workflows/hil-integration-esp-idf.yml
+++ b/.github/workflows/hil-integration-esp-idf.yml
@@ -82,7 +82,7 @@ jobs:
     timeout-minutes: 30
 
     container:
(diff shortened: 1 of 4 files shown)


sam-golioth:
@mniestroj Is this one ready for review? Just checking because there are no requested reviewers.

mniestroj:
> @mniestroj Is this one ready for review? Just checking because there are no requested reviewers.

Yes, it is. Didn't assign since tests were still failing. I don't see why it would be due to those images update, so we can probably move forward with it.

sam-golioth (approved):

PR #809, opened 2025-07-17
location: move location support to net_info

hasheddan:
Updates the Golioth Location functionality to instead construct network information messages that are delivered to Golioth via stream. This network info may be converted into position information using Golioth Location, but a Pipeline must be configured to do so.

PR #810, opened 2025-07-18
src: replace Zephyr style log with cross-platform macros

szczys:
Replace LOG_XXX() with GLTH_LOGX() calls. This brings files in line with the standard for cross-platform files. Even though these are currently Zephyr-specific files, this ensures that these may be enabled/disabled using CONFIG_GOLIOTH_DEBUG_LOG in the same way as all other files.

resolves https://github.com/golioth/firmware-issue-tracker/issues/816

### DIFF
--- a/src/coap_client_zephyr.c
+++ b/src/coap_client_zephyr.c
@@ -332,7 +332,7 @@ static int golioth_coap_cb(struct golioth_req_rsp *rsp)
     {
         if (!golioth_sys_timer_reset(client->keepalive_timer))
         {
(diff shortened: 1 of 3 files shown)


sam-golioth:
Let's drop the changes from the location files. #809 is going to delete them so it will just cause conflicts. @hasheddan can you update the remaining `LOG_XXX` calls in those files in your PR? Or we can open another one after that merges.

sam-golioth (approved):

PR #818, opened 2025-08-07
Fix undefined reference warnings when debug logging is turned off

sam-golioth:
These errors are due to using snprintf() but not directly including stdio.h. In lightdb_state.c and ota.c we now include stdio.h. In settings.c we rework logging to remove the snprintf call, which also recovers some resources when logging is disabled.

### DIFF
--- a/src/lightdb_state.c
+++ b/src/lightdb_state.c
@@ -4,7 +4,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 #include <assert.h>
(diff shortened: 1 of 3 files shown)


sam-golioth:
@cdwilson FYI

cdwilson:
@sam-golioth just tried this out, it fixes the warnings I was seeing previously. Thx!

hasheddan (approved):

PR #821, opened 2025-08-13
gateway: update pouch uplink / donwlink path

hasheddan:
Updates the path used for pouch uplink / downlink.


sam-golioth:
Let's wait on this one until after we upstream the patches from the gateway repo. It will cause conflicts and we'll need to redo all the patches, might as well put that effort into upstreaming.

sam-golioth (approved):

PR #829, opened 2025-09-19
ota: periodically poll for manifest

sam-golioth:


src/Kconfig
```diff
@@ -32,6 +32,14 @@ config GOLIOTH_OTA_MAX_NUM_COMPONENTS
     help
         Maximum number of components in an OTA manifest.
 
+config GOLIOTH_OTA_MANIFEST_OBSERVE_PERIOD_S
```
hasheddan, line 35: nitpick: having `OBSERVE` in the name kind of feels like it is related to observations specifically, and though this works alongside observations for OTA, the polling is kind of an alternative. Curious how you would feel about `GOLIOTH_OTA_MANIFEST_POLL_PERIOD_S` or `GOLIOTH_OTA_MANIFEST_CHECK_PERIOD_S` or similar? I could imagine us eventually having a `y/n` observe and then a polling interval.

src/ota.c
```diff
@@ -88,10 +96,49 @@ const struct golioth_ota_component *golioth_ota_find_component(
...
+    if (!manifest_timer_initialized && CONFIG_GOLIOTH_OTA_MANIFEST_OBSERVE_PERIOD_S > 0)
+    {
+        struct golioth_timer_config cfg =
+        {
+            .name = "ota_manifest_timer",
+            .expiration_ms = CONFIG_GOLIOTH_OTA_MANIFEST_OBSERVE_PERIOD_S * 1000,
+            .fn = ota_manifest_timer_expiry,
+            .user_arg = &manifest_timer_arg,
+        };
+        manifest_timer = golioth_sys_timer_create(&cfg);
```
hasheddan, line 132: My first reaction was that it is a bit out of place for this timer / polling logic to be handled in the OTA client methods and coupled to establishing an observation. However, in thinking about it more, if we elevate the concept of "observing" beyond the CoAP implementation of observations and instead make observe mean the discussed "polling + short circuit" behavior then I think managing the timer within the OTA client makes sense 👍🏻

src/Kconfig
hasheddan: Keeping my comment above for context, but after doing more thinking (see other thread), I like the idea of making a Golioth SDK concept of "observations" and using that here 👍🏻

src/ota.c
```diff
@@ -88,10 +96,49 @@ const struct golioth_ota_component *golioth_ota_find_component(
...
+        manifest_timer_initialized = true;
+    }
+
+    if (manifest_timer_initialized)
+    {
+        golioth_sys_timer_reset(manifest_timer);
+    }
+
     uint8_t token[GOLIOTH_COAP_TOKEN_LEN];
     golioth_coap_next_token(token);
```
hasheddan, line 144: Can't comment below this line, but I think it could make sense to establish the observation first, check the error, then start the timer if establishing the observation was successful. From a caller's perspective, it would be confusing IMO to attempt to observe the manifest, have an error returned, but then have a periodic timer running in the background to fetch the manifest. The downside to attempting to establish the observation first is that you would be blocked from polling until you could establish an observation. However, if you are unable to establish an observation when retrying it is unlikely the polling get is going to work either.

src/ota.c
```diff
@@ -88,10 +96,49 @@ const struct golioth_ota_component *golioth_ota_find_component(
...
+    struct manifest_get_args *get_args = user_arg;
+
+    enum golioth_status status = golioth_ota_get_manifest_async(get_args->client,
+                                                                get_args->cb,
+                                                                get_args->arg);
+
+    if (GOLIOTH_OK != status)
+    {
+        GLTH_LOGE(TAG, "Periodic OTA manifest fetch failed");
+    }
```
hasheddan, line 110: Because observation establishment / cancellation is handled internally to the client, we can't easily cancel this timer, so I expect that if the client is stopped then we'll see this error logged whenever the timer fires? I think that is okay for the time-being, but if we move more towards the harmonic polling you've described in the future I suspect we could move to a single internally managed timer and avoid issuing requests when the client is stopped.

hasheddan (commented):
Awesome to see @sam-golioth! There are a lot of words in the comments below, but really the only thing I'm actually proposing in this PR is swapping the order of observation establishment and timer initialization 😅 The rest ended up just being musing and some future alignment.

src/ota.c
sam-golioth: I was thinking that it would be good to have the polling timer as a backup to observations, but you're right that neither is likely to succeed without the other. And thinking about it more, really this is a "never established connection" issue, and handling that should be outside the purview of the OTA module.

I'll swap the order as you suggest 👍

src/Kconfig
sam-golioth: Yep, that's exactly the intention, but I spent a good deal of time trying to think of a better verbiage and I'm still not totally satisfied, so I'm very open to suggestions. Both "observation" and "polling" are incomplete on their own, and things like "worst case latency" or something like that to describe the period isn't entirely accurate either, as any individual GET operation could still fail.

sam-golioth:
@hasheddan I believe I addressed all your feedback. I also changed the name of the function that registers manifest observations to reflect the new terminology.

PR #830, opened 2025-09-30
zephyr: wifi: Add SAE support

morse-cmistry:
Currently we either use PSK security or none based on whether or not a password is supplied.
This change adds functionality to set the wifi key type in advance, allowing the user to set SAE.

### DIFF
--- a/examples/zephyr/common/wifi.c
+++ b/examples/zephyr/common/wifi.c
@@ -147,6 +147,8 @@ static uint8_t wifi_ssid[WIFI_SSID_MAX_LEN];
 static size_t wifi_ssid_len;
 static uint8_t wifi_psk[WIFI_PSK_MAX_LEN];
 static size_t wifi_psk_len;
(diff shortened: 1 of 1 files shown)


sam-golioth:
Thanks for the contribution @morse-cmistry! You're timing is pretty good, we're in the [process](https://github.com/golioth/golioth-firmware-sdk/pull/832) of moving to using Zephyr's [WiFi Credentials](https://docs.zephyrproject.org/latest/connectivity/networking/api/wifi_credentials.html) library and deprecating our own implementation. This will allow you to add credentials via the built-in `wifi` [shell](https://docs.zephyrproject.org/latest/connectivity/networking/api/wifi_credentials.html#shell-commands), which offers much more flexibility for the kinds of networks and security schemes it supports. Once we have that in, you'll be able to use SAE keys.

morse-cmistry:
@sam-golioth thanks for letting me know.

PR #832, opened 2025-10-02
examples: zephyr: use Zephyr's WiFi credentials library

sam-golioth:


examples/zephyr/stream/README.md
```diff
@@ -37,13 +37,14 @@ uart:-$ kernel reboot cold
 
 #### WiFi Configuration
```
hasheddan, line 38: May make sense to move this out to a common location that all examples can reference in the future to avoid having to update n times whenever something changes.

hasheddan (approved):

examples/zephyr/stream/README.md
sam-golioth: Couldn't agree more

PR #906, opened 2026-06-02
CI: Clean up auto-generated main packages

szczys:
Clean up auto-generated `main` packages with a known version string pattern that have been orphaned by a CI run.

I've already run this a few times to clean up about 21 pages of orphaned artifacts. Adding this to the workflow will keep any future leaks cleaned up on a nightly basis.

This resolves the failing recurring OTA test which were cause by the REST API returning a limited number of `main` artifacts that didn't include the CI target artifacts do to the number of orphaned packages in the project.


hasheddan (approved):
Thanks @szczys!

PR #908, opened 2026-06-18
Port: Zephyr: Fix thread stack handling during create and destroy

### LINKED ISSUE #907: Zephyr port: thread-stack pool not recycled + off-by-one bounds check causes hard fault on client recreate
### Golioth Firmware SDK Commit Hash

c3845a4563fc09119f510340acc3a40f27b5ff0c

### Platform/OS

Zephyr v3.2.1

### Hardware Target

_No response_

### What host OS are you using?

_No response_

### Current Behavior

I hit this when implementing connection recovery logic. I called `golioth_client_destroy` and then `golioth_client_create`. After doing this twice the third cycle resulted in a memory usage hard fault.

After looking into the Golioth SDK source I noticed that there are two bugs in **port/zephyr/golioth_sys_zephyr.c** which contribute to causing this:

1) In `golioth_sys_thread_destroy` the thread gets aborted and freed but there is no reuse of entries in `golioth_thread_stacks`. These get exhausted as `golioth_sys_thread_destroy` is called.
2) In `stack_alloc` there is an off-by-one bug in the check of the thread stack slot index. When the index reaches the array size it is out of bounds but will pass the check. As a result the function returns an invalid stack pointer, which triggers the hard fault I see when `k_thread_create` is subsequently called on it.

Here is the location of the off-by-one bug: https://github.com/golioth/golioth-firmware-sdk/blob/c3845a4563fc09119f510340acc3a40f27b5ff0c/port/zephyr/golioth_sys_zephyr.c#L225-L228

### Expected Behavior

My expectation was that it would be safe to repeatedly call `golioth_client_destroy` and then `golioth_client_create`.

### Steps To Reproduce

_No response_

### Logs, console output, or any kind of debug information

```shell

```

szczys:
Update the Zephyr port to track stack thread thread allocation via the index of the stack in the pool. By storing this pool index during creation in the golioth_thread struct, the destroy function can mark the slot as available for future use.

resolves: #907

### DIFF
--- a/port/zephyr/golioth_sys_zephyr.c
+++ b/port/zephyr/golioth_sys_zephyr.c
@@ -208,6 +208,7 @@ void golioth_sys_timer_destroy(golioth_sys_timer_t gtimer)
 struct golioth_thread
 {
     struct k_thread thread;
+    int stack_idx;
     k_tid_t tid;
     golioth_sys_thread_fn_t fn;
     void *user_arg;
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
Thanks for addressing this @szczys!

PR #910, opened 2026-06-23
zephyr: sample: common: check for wifi ready before connect

szczys:
Add wifi_iface_ready() to detect when the WiFi interface is ready to connect. Use it to guard the connect operation so that devices that need more time to bring up the interface (eg: nrf70) do not immediately fail to connect because the interface is still spinning up.

When wifi_iface_ready() returns false, the guard enqueues a disconnect event. Because the disconnect event fired during the connecting state, it triggers a retry after 1000ms using the wifi_state_change_map already in place.

### DIFF
--- a/examples/zephyr/common/wifi.c
+++ b/examples/zephyr/common/wifi.c
@@ -168,9 +168,38 @@ static void wifi_mgmt_connecting_update(struct wifi_manager_data *wifi_mgmt, boo
     }
 }
 
+static bool wifi_iface_ready(struct wifi_manager_data *wifi_mgmt)
+{
+    struct wifi_iface_status status = {0};
+    int err;
+
+    err = net_mgmt(NET_REQUEST_WIFI_IFACE_STATUS, wifi_mgmt->iface, &status, sizeof(status));
+    if (err)
+    {
+        LOG_DBG("WiFi iface status not ready yet: %d", err);
+        return false;
+    }
+
+    switch (status.state)
+    {
+        case WIFI_STATE_INTERFACE_DISABLED:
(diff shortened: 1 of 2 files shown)


examples/zephyr/common/wifi.c
```diff
@@ -168,10 +168,48 @@ static void wifi_mgmt_connecting_update(struct wifi_manager_data *wifi_mgmt, boo
...
+        case WIFI_STATE_AUTHENTICATING:
+        case WIFI_STATE_4WAY_HANDSHAKE:
+        case WIFI_STATE_GROUP_HANDSHAKE:
+        case WIFI_STATE_COMPLETED:
+            return true;
+        case WIFI_STATE_INTERFACE_DISABLED:
+        case WIFI_STATE_UNKNOWN:
+        default:
+            LOG_DBG("WiFi iface not ready, state=%d", status.state);
+            return false;
```
hasheddan: Do we want the default to be `false`, which will prevent us from ever connecting successfully on receiving an unknown state, or would it make more sense to return `true` in all cases other than `WIFI_STATE_INTERFACE_DISABLED` and `WIFI_STATE_UNKNOWN`?

szczys: Great suggestion, thank you! I've updated the commit to work in this way.

szczys:
While addressing PR feedback I encountered another issue where a crash happens when net_connect() is called but no interface is available. I added a commit to log and return, preventing this crash which in my case was completely blocking all logging prior to the change.

trond-snekvik (approved):

hasheddan (approved):

PR #912, opened 2026-06-25
CI: fix how logs are filtered for the ota_reason_and_state_test

szczys:
stabilize the flaky test by:
- validating based on log message content and not absolute number of logs
  - if validation fails, keep polling for new messages until timeout
- filter received logs by reason code and fail on duplicates

### DIFF
--- a/tests/hil/tests/ota/test_ota.py
+++ b/tests/hil/tests/ota/test_ota.py
@@ -119,15 +119,33 @@ async def verify_component_values(board, info):
     assert None is not await board.wait_for_regex_in_line(f'component.hash: {b_info["digests"]["sha256"]["digest"]}', timeout_s=2)
 
 def verify_dfu_status_logs(logs):
-    assert len(logs) == len(golioth_ota_reason)
+    # Filter to only the DFU status-report logs emitted by test_reason_and_state.
+    status_logs = [
+        log for log in logs
+        if log.metadata.get('reasonCode') is not None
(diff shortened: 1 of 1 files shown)


trond-snekvik (approved):

tests/hil/tests/ota/test_ota.py
```diff
@@ -119,15 +119,33 @@ async def verify_component_values(board, info):
     assert None is not await board.wait_for_regex_in_line(f'component.hash: {b_info["digests"]["sha256"]["digest"]}', timeout_s=2)
 
 def verify_dfu_status_logs(logs):
-    assert len(logs) == len(golioth_ota_reason)
+    # Filter to only the DFU status-report logs emitted by test_reason_and_state.
```
hasheddan, line 122: Not a high priority at the moment, but would be great to switch to OTA events API in the future (requires update to Python tools)

hasheddan (approved):
Thanks @szczys!

szczys:
I was confused by the failure of the Firebase workflow. It seems that when you don't change one of the files used in the doxygen build (a header or a source file) the doxygen build will not be unique and a second workflow run will fail because the doxygen build is the same version as the previous deployment. [That's the case here](https://github.com/golioth/golioth-firmware-sdk/actions/runs/28198246241/job/83530694726?pr=912) so we are good to override this requrement to merge this PR.

mniestroj (approved):

PR #914, opened 2026-06-29
CI: Pin all actions and containers and grant write permissions where neccessary

szczys:
- Upgrade all actions to latest release and pin by sha
- Pin all containers to digest sha (do not change any versions)
- Add write permissions to workflows that require them

The Golioth Org default has changed to only grant read permission. This PR grants write permissions to the workflows that need them.

Relates to: https://github.com/golioth/firmware-issue-tracker/issues/958

### DIFF
--- a/.github/workflows/ci-clean-project.yml
+++ b/.github/workflows/ci-clean-project.yml
@@ -11,10 +11,10 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Setup python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: '3.11' # >=3.11 for datetime.fromisoformat()
 
--- a/.github/actions/safe-upload-artifacts/action.yml
+++ b/.github/actions/safe-upload-artifacts/action.yml
@@ -65,7 +65,7 @@ runs:
         done
 
     - name: Upload artifacts
-      uses: actions/upload-artifact@v6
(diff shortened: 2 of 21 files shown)


szczys:
The first doxygen workflow run failed becasue we were missing `checks: write`. The next one failed because the same versionis already deployed (no header or source file changes in this PR). I believe this is expected and can be overridden to merge.

hasheddan (approved):
Thanks @szczys!

.github/workflows/ci-clean-project.yml
```diff
@@ -11,10 +11,10 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
```
mniestroj: Do we want to go to 7.0.0 already?

mniestroj (approved):
Just a minor comment about using 7.0.0 version of actions/checkout.

.github/workflows/ci-clean-project.yml
szczys: Thanks for catching this! I've updated these to v7 and will push the changes once our firebase is ready for another test.

szczys:
@hasheddan I asked for another review to ensure the workaround in this commit looks okay to you:
https://github.com/golioth/golioth-firmware-sdk/pull/914/changes/ba9e941ae52e822e59afd3495d19ae6fcde431db

EDIT: oof, posted this too early, the sample/integration tests are still getting blocked by this. Fixing now...

I added a tracking issue for updating this in the future:
https://github.com/golioth/firmware-issue-tracker/issues/1027

hasheddan (approved):
Thanks @szczys -- looks like a reasonable short term fix 👍🏻

szczys:
Okay, turns out we don't need a workaround. I was trying to pin the version of firebase-tools but this absolute rockstar figured out that pinning an earlier version of node24 prevents the failures:

https://github.com/firebase/firebase-tools/issues/10716#issuecomment-4800040219

I've implemented this and will merge once tests finish.

PR #920, opened 2026-07-20
Fix  memory leak in golioth_ota_download_components()

### LINKED ISSUE #919: OTA: golioth_ota_download_component leaks its ctx on a failed request-issue (+ unchecked malloc)
### Summary

`golioth_ota_download_component()` (`src/ota.c`) has two heap issues on its request-issue path:

1. **Unchecked `golioth_sys_malloc`** — the allocation is dereferenced without a NULL check.
2. **The wrapper `ctx` is leaked when the blockwise GET fails to enqueue** — `ctx` is freed only by the end-callback wrapper, which never runs if `golioth_blockwise_get()` returns an error before the transfer starts.

Observed on **v0.22.0**.

### Details

`src/ota.c`, `golioth_ota_download_component()` (~lines 498–515):

```c
enum golioth_status golioth_ota_download_component(struct golioth_client *client,
                                                   const struct golioth_ota_component *component,
                                                   uint32_t block_idx,
                                                   ota_component_block_write_cb block_cb,
                                                   ota_component_download_end_cb end_cb,
                                                   void *arg)
{
    struct ota_component_blockwise_ctx *ctx =
        golioth_sys_malloc(sizeof(struct ota_component_blockwise_ctx));
    ctx->component = component;   // (1) unchecked -> NULL-deref if malloc fails
    ctx->block_cb = block_cb;
    ctx->end_cb = end_cb;
    ctx->arg = arg;

    return golioth_blockwise_get(client, "", component->uri,
                                 GOLIOTH_CONTENT_TYPE_OCTET_STREAM, block_idx,
                                 ota_component_write_cb_wrapper,
                                 ota_component_download_end_cb_wrapper,
                                 ctx);   // (2) if this returns != OK, ctx leaks
}
```

`ctx` is freed **only** in `ota_component_download_end_cb_wrapper()` (`src/ota.c`, `golioth_sys_free(ctx)` ~line 489), which runs only when the transfer reaches its end callback. But `golioth_blockwise_get()` has early error returns *before* it takes ownership / starts the transfer — e.g. `GOLIOTH_ERR_NULL` (`src/coap_blockwise.c:508`), `GOLIOTH_ERR_MEM_ALLOC` (`:514`), `GOLIOTH_ERR_FAIL` (`:520`), and a full request mailbox (`golioth_mbox_try_send` → `GOLIOTH_ERR_QUEUE_FULL`). On any of those the end callback never fires, so the `ctx` allocated here is never freed.

(For contrast, `golioth_blockwise_get()` itself *does* check its own `golioth_sys_malloc` at `coap_blockwise.c:511-514`; the OTA wrapper just doesn't.)

### Impact

The leak is small (`sizeof(struct ota_component_blockwise_ctx)`, ~16 B on 32-bit) but recurring: an application that re-issues the component download to resume after a transient failure leaks one `ctx` per failed issue, and `QUEUE_FULL` from request-mailbox congestion is quite plausible mid-OTA. On a RAM-constrained target (we hit this on an ESP32-C6) it accumulates across a flaky OTA. The unchecked malloc is a latent NULL-deref under heap pressure.

### Suggested fix

```c
struct ota_component_blockwise_ctx *ctx =
    golioth_sys_malloc(sizeof(struct ota_component_blockwise_ctx));
if (ctx == NULL)
{
    return GOLIOTH_ERR_MEM_ALLOC;
}
ctx->component = component;
ctx->block_cb = block_cb;
ctx->end_cb = end_cb;
ctx->arg = arg;

enum golioth_status status =
    golioth_blockwise_get(client, "", component->uri,
                          GOLIOTH_CONTENT_TYPE_OCTET_STREAM, block_idx,
                          ota_component_write_cb_wrapper,
                          ota_component_download_end_cb_wrapper, ctx);
if (status != GOLIOTH_OK)
{
    golioth_sys_free(ctx);
}
return status;
```

Happy to send a PR if that's useful.

szczys:
As reported by @rlinus in #919, the failure path in goliioth_ota_download_components() returns without freeing the allocated context. This PR follows his suggested fix to test that status code and free on failure, as well as checking for NULL after the initial malloc.

A review of treewide golioth_sys_malloc() calls was conducted to ensure memory is freed correctly and a similar leak was fixed in the ota HIL test. Instances where free() was called where golioth_sys_free() should have been called were also fixed, which is a correction of the API use but does not result in any changed behavior.

resolves #919

### DIFF
--- a/src/coap_blockwise.c
+++ b/src/coap_blockwise.c
@@ -293,7 +293,7 @@ enum golioth_status golioth_blockwise_post(struct golioth_client *client,
     golioth_sys_sem_destroy(ctx.sem);
 
 finish_with_block_buffer:
-    free(ctx.block_buffer);
+    golioth_sys_free(ctx.block_buffer);
 
 finish:
     return status;
--- a/src/net_info.c
+++ b/src/net_info.c
@@ -36,7 +36,7 @@ enum golioth_status golioth_net_info_destroy(struct golioth_net_info *info)
         return GOLIOTH_ERR_NULL;
     }
 
-    free(info);
+    golioth_sys_free(info);
     return GOLIOTH_OK;
 }
 
--- a/src/ota.c
+++ b/src/ota.c
@@ -448,6 +448,7 @@ static void ota_component_download_end_cb_wrapper(struct golioth_client *client,
 
     ctx->end_cb(status, coap_rsp_code, ctx->component, block_idx, ctx->arg);
 
+    /* allocated in golioth_ota_download_component(), freed here */
     golioth_sys_free(ctx);
 }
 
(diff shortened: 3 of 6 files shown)


mniestroj (approved):

hasheddan (approved):
Thanks @szczys and @rlinus!

trond-snekvik (approved):

PR #925, opened 2026-08-12
Add support for the WiFi variant of frdm_rw612

szczys:
Add optional conf files for using frdm_rw612 with WiFi (instead of Ethernet). Add tests to all samples for this variant.

This PR also adds an erase step before the fw_update test runs. This is necessary to ensure the mcuboot flags from a previous (failed) run are wiped out before starting a new test.

resolves https://github.com/golioth/firmware-issue-tracker/issues/1038

### DIFF
--- a/examples/zephyr/fw_update/boards/frdm_rw612_wifi.conf
+++ b/examples/zephyr/fw_update/boards/frdm_rw612_wifi.conf
@@ -1,0 +1,33 @@
+# Extra configuration file for using NXP frdm_rw612 with WiFi instead of Ethernet
+#   To use, add as extra conf file:
+#   west build -p -b frdm_rw612 examples/zephyr/fw_update \
+#             --sysbuild                                  \
+#             --extra-conf boards/frdm_rw612_wifi.conf
+#
+CONFIG_WIFI=y
+CONFIG_WIFI_NXP=y
+CONFIG_GOLIOTH_SAMPLE_WIFI=y
+
+# Runtime provisioning for WiFi credentials
+CONFIG_WIFI_CREDENTIALS=y
+CONFIG_NET_L2_WIFI_SHELL=y
+CONFIG_SHELL_STACK_SIZE=2560
+
+# Wifi Configuration
+CONFIG_NET_L2_ETHERNET=y
+CONFIG_NET_PKT_RX_COUNT=36
+CONFIG_NET_PKT_TX_COUNT=36
+CONFIG_NET_BUF_RX_COUNT=40
+CONFIG_NET_BUF_TX_COUNT=40
+CONFIG_NET_BUF_DATA_SIZE=1744
+
+# Handle WiFi Networks with Multiple DNS Servers
+CONFIG_DNS_RESOLVER_MAX_SERVERS=3
+
+# Disable Ethernet
+CONFIG_ETH_NXP_ENET=n
+CONFIG_PHY_MICROCHIP_KSZ8081=n
+CONFIG_NET_CONFIG_AUTO_INIT=n
+CONFIG_NET_CONFIG_NEED_IPV4=n
+
+CONFIG_ZVFS_POLL_MAX=4
--- a/examples/zephyr/fw_update/pytest/conftest.py
+++ b/examples/zephyr/fw_update/pytest/conftest.py
@@ -1,4 +1,13 @@
 import pytest
+import sys
+import west.configuration
+from pathlib import Path
+from twister_harness.helpers.domains_helper import get_default_domain_name
+
+WEST_TOPDIR = Path(west.configuration.west_dir()).parent
+
+sys.path.insert(0, str(WEST_TOPDIR / 'zephyr' / 'scripts' / 'west_commands'))
+from runners.core import BuildConfiguration
 
 UPDATE_VERSION = '255.8.9'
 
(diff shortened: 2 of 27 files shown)


szczys:
The wifi variant of the tests all passed. There are some issues that I'm tracking but I don't think they should hold up this PR:
1. The yellow results are all failed Ethernet. I believe I already have a fix for this on another branch and I don't believe these results have anything to do with the addition of WiFi here
2. I expected the fw_update test to have run but it is not included in these results (edit: this is resolved now)

image

szczys:
I added two commits to this and [tested in isolation](https://github.com/golioth/golioth-firmware-sdk/actions/runs/33654479938/) with good success:
1. Checking the link before DHCP should deliver better stability. I did see one of the hello tests fail because of no link but this change gave the run several minutes to retry link negotiation where before it would have failed immediately so this was better IMO.
2. A failed step was resulting in the upload of stale run artifacts. Removing the twister-out directory earlier resolves this.

examples/zephyr/fw_update/boards/frdm_rw612_wifi.conf
```diff
@@ -0,0 +1,30 @@
+# Extra configuration file for using NXP frdm_rw612 with WiFi instead of Ethernet
+#   To use, add as extra conf file:
+#   west build -p -b frdm_rw612 <path_to>/examples/zephyr/http_client -- -DEXTRA_CONF_FILE=boards/frdm_rw612_wifi.conf
```
mniestroj: Maybe we should drop `<path_to>/examples/zephyr/http_client`, so that we can copy-paste that config safely in the future.


Additionally, we can use:
```
west build -p -b frdm_rw612 ... --extra-conf boards/frdm_rw612_wifi.conf
```
instead of:
```
west build -p -b frdm_rw612 ...  -- -DEXTRA_CONF_FILE=boards/frdm_rw612_wifi.conf
```
This is just FYI, I do not have a strong preference on that.

mniestroj: Oh, it seems that you have fixed that, but in subsequent commit. I was reviewing commit by commit, so that is why I have noticed that.

examples/zephyr/fw_update/boards/frdm_rw612_wifi.conf
```diff
@@ -1,9 +1,10 @@
 # Extra configuration file for using NXP frdm_rw612 with WiFi instead of Ethernet
 #   To use, add as extra conf file:
-#   west build -p -b frdm_rw612 <path_to>/examples/zephyr/http_client -- -DEXTRA_CONF_FILE=boards/frdm_rw612_wifi.conf
+#   west build -p -b frdm_rw612 <path_to_example> -- -DEXTRA_CONF_FILE=boards/frdm_rw612_wifi.conf
 #
 CONFIG_WIFI=y
 CONFIG_WIFI_NXP=y
+CONFIG_GOLIOTH_SAMPLE_WIFI=y
```
mniestroj, line 9: This should probably land in previous commit as well.

examples/zephyr/fw_update/pytest/conftest.py
```diff
@@ -8,7 +8,11 @@ def anyio_backend():
 
 @pytest.fixture(scope="session")
 async def target_package(request):
-    return request.config.getoption('hil_board')
+    hil_board = request.config.getoption('hil_board')
+    build_dir = request.config.getoption('build_dir', '')
+    if '.wifi' in build_dir:
+        return f'{hil_board}_wifi'
+    return hil_board
```
mniestroj: That is a bit hacky, but I do not have better solution for now. Let's have that as it is.

examples/zephyr/fw_update/sample.yaml
```diff
@@ -36,6 +36,7 @@ tests:
     timeout: 720
     extra_configs:
       - CONFIG_GOLIOTH_SAMPLE_TWISTER_TEST=y
+      - CONFIG_GOLIOTH_FW_UPDATE_PACKAGE_NAME="frdm_rw612_wifi"
```
mniestroj, line 39: Should we maybe squash that with previous commits as well?

examples/zephyr/fw_update/pytest/conftest.py
mniestroj: Actually... maybe we should fetch `CONFIG_GOLIOTH_FW_UPDATE_PACKAGE_NAME` configuration in pytest.

examples/zephyr/fw_update/boards/frdm_rw612_wifi.conf
szczys: This is just some sloppy rebasing on my part. I apologize that you had to dig through these but I'll get them cleaned up. Thanks!

examples/zephyr/fw_update/pytest/conftest.py
szczys: At first I thought you might be right but then I think we need to parse the `.config` file, is that right?

That's a little wonky. We would also need to locate the app build directory since we're using `--sysbuild`. I wonder if it's cleaner to leave it as is?

examples/zephyr/fw_update/sample.yaml
szczys: I think the work to add the wifi variant to the fw_update test should be its own commit. I've separated out the parts for that and left them in this commit but moved the workflow changes for fetching the nxp blobs to the previous commit. This should be a clean an logical separation now.

mniestroj (approved):

examples/zephyr/fw_update/pytest/conftest.py
mniestroj: Yes, we would need to parse `.config` of the app build directory. We do something similar in https://github.com/golioth/golioth-firmware-sdk/blob/eb99bc096c5137fc5f1af78931222ac705f9e527/examples/zephyr/certificate_provisioning/pytest/conftest.py#L15-L38 and https://github.com/golioth/golioth-firmware-sdk/blob/eb99bc096c5137fc5f1af78931222ac705f9e527/examples/zephyr/certificate_provisioning/pytest/test_sample.py#L31-L34

Your call whether we apply that.

PR #927, opened 2026-08-27
Zephyr: Resolve non-working board targets on certificate auth samples

szczys:
The mbedtls update to v3.6.5 broke certificate authentication for:
- frdm_rw612
- nrf52840dk

Kconfig settings for the frmd_rw612 have been adjusted to resolve the allocation errors and handshake timeout that was observed and the board once again works with the certificate_provisioning sample.

Support for the nrf52840dk has been removed from certificate_provisioning and certificate_rotation to resolve this issue. An mbedtls change replaced mbedtls_mpi_inv_mod() in favor of constant time mbedtls_mpi_gcd_modinv_odd(). The nrf52840dk on Zephyr uses software-based crypto and is unable to finish the handshake before the max 60s timeout for the connection is reached. This board would very likely pass if it were built using NCS with the cryptocell enabled but that work is beyond the scope of this PR.


examples/zephyr/certificate_provisioning/boards/frdm_rw612.conf
```diff
@@ -3,5 +3,5 @@ CONFIG_NET_DHCPV4=y
 CONFIG_BUILD_OUTPUT_HEX=y
 
 # Needed for certificate auth
-CONFIG_NET_PKT_TX_COUNT=10
-CONFIG_NET_SOCKETS_CONNECT_TIMEOUT=6000
+CONFIG_NET_BUF_DATA_SIZE=512
+CONFIG_NET_SOCKETS_CONNECT_TIMEOUT=12000
```
mniestroj-ai: Please consider enabling `MBEDTLS_ECP_NIST_OPTIM` option, as we did in pouch: https://github.com/golioth/pouch/commit/433d9e83ab627aa9af1fbd6efb0dbc46bec0a47b. This will probably speed-up software-based handshake, so that timeout increase will not be needed.

szczys: Thanks, that's helpful. I've added a commit to use this symbol instead of bumping the timeout.

mniestroj (approved):
Awesome work on that, thanks!

